PR #45, opened 2015-01-29
added output format options

chaudum:
- raw json output
- json output that maps column name
- coloured output for interactive usage

### DIFF
--- a/src/crate/crash/command.py
+++ b/src/crate/crash/command.py
@@ -31,6 +31,16 @@ import os
 import sys
 import select
 
+from colorama import init, Fore, Style
+init(autoreset=True)
+
+from pygments import highlight
+from pygments.lexers.data import JsonLexer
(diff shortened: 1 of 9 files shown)


src/crate/crash/command.py
```diff
@@ -61,6 +68,26 @@
...
                         padding=1, with_header_hide=None)
 
+class ColorLog(object):
+
+    def __init__(self, isatty=None):
+        self.pretty = isatty is None and sys.stdout.isatty() or isatty
+
+    def log(self, content, color, bold=False):
+        if self.pretty:
+            print('\033[{2};{1}m{0}\033[0m'.format(content, str(color), int(bold)))
```
mfussenegger: will this work cross platform? otherwise you could use colorama or termcolor?

mfussenegger:
ack :+1:

mfussenegger:
ack2

PR #60, opened 2015-04-08
fixed misconfigured logging

chaudum:
that caused warning logs from the crate-python library being
printed to stderr


pustovit:
ack

PR #62, opened 2015-04-13
Fixed multilining command interpretation

pustovit:


src/crate/crash/command.py
```diff
@@ -368,6 +368,8 @@ def _try_exec_cmd(self, line):
             if message:
                 self.logger.info(message)
             return True
+        else:
+	    self.logger.critical('Command unknown')
```
chaudum: better error message:

Unknown command '<command>.' Type `\q` for a full list of available commands.

src/crate/crash/command.py
```diff
@@ -465,9 +467,10 @@ def _create_default_layout(lines_ref):
 def is_multiline(doc):
     if not doc.text:
         return False
+    if doc.text.startswith('\\'):
+	return False
```
chaudum: `cr> \? ;` is still a valid command in your case, but should not be ;)

CHANGES.txt
```diff
@@ -7,6 +7,8 @@ Unreleased
 
  - added ``--version`` to cli interface to show crash version and exit
 
+ - Attempting to use unknown command now results in an error
```
chaudum: move to top of list

chaudum:
did you run tests? there are problems with indentation
`bin/tox` does the job for all supported python versions

chaudum:
ack

PR #68, opened 2015-04-30
add 'dynamic' output format

mfussenegger:


chaudum:
changelog entry is missing. otherwise ack

chaudum:
ack

PR #73, opened 2015-07-09
added multiline support for tabular output format

chaudum:
based on https://bitbucket.org/astanin/python-tabulate/issues/28/feature-request-multiline-cells-and


mfussenegger:
ack

PR #75, opened 2015-08-12
updated crate-python to 0.13.4

chaudum:


src/crate/crash/test_command.py
```diff
@@ -166,7 +166,7 @@ def test_multiple_hosts(self):
                     exception_code = e.code
                     self.assertEqual(exception_code, 0)
                     output = output.getvalue()
-                    self.assertTrue("| http://127.0.0.1:44209     | crate     | 0.49.2  | TRUE      | OK" in output, output)
+                    self.assertTrue("| http://127.0.0.1:44209     | crate     | 0.50.3  | TRUE      | OK" in output, output)
```
bogensberger: Could you change that assertion so it's not necessary to always update that when upgrading crate?

bogensberger:
ack

PR #76, opened 2015-08-14
R/check command

kovrus:


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+ - Added the check command
```
chaudum: please be more precise about the feature

src/crate/crash/command.py
```diff
@@ -236,7 +240,8 @@ def __init__(self, connection=None, error_trace=False,
             'c': self._connect,
             'format': self._switch_format,
             'connect': self._connect,
-            'dt': self._show_tables
+            'dt': self._show_tables,
+            'check': self._check
```
chaudum: please add trailing `,` after last item

src/crate/crash/command.py
```diff
@@ -402,6 +441,8 @@ def _connect(self, server):
             self.logger.critical('CONNECT ERROR')
         else:
             self.logger.info('CONNECT OK')
+        if connected:
```
chaudum: the `connected` is not necessary, because you an put the `check()` call into the `else` block

src/crate/crash/command.py
```diff
@@ -369,6 +374,38 @@ def _show_tables(self, *args):
                       where schema_name not in ('sys','information_schema')""")
 
     @noargs_command
```
chaudum: decorator only required for commands that are called in the `_try_exec_cmd` method

src/crate/crash/command.py
```diff
@@ -369,6 +374,38 @@ def _show_tables(self, *args):
...
+            'rowcount': cur.rowcount
+        }
+        checks = cur.fetchall()
+        if len(checks):
+            self.pprint(checks, [c[0] for c in cur.description])
+            tmpl = '{rowcount} CHECK{s} FAILED'
+            self.logger.critical(tmpl.format(**print_vars))
+        else:
+            tmpl = 'CLUSTER CHECK OK'
+            self.logger.info(tmpl.format(**print_vars))
```
chaudum: no need to use a template. just log the string

src/crate/crash/command.py
```diff
@@ -369,6 +374,38 @@ def _show_tables(self, *args):
...
+        return True
+
+    @noargs_command
+    def _check(self, *args):
+        """ print failed cluster checks """
+        if self.connection.lowest_server_version >= CHECK_MIN_VERSION:
+            self.check()
+        else:
+            tmpl = '\nCrate {version} version does not support the cluster "check" command'
```
chaudum: Crate {version} ~~version~~ does not support ...

chaudum: also check if the command was called via `\check` or on initial connect,
there should not be a warning on initial connect and Crate does not support cluster checks

chaudum:
ack. wait for green travis

docs/cli.txt
```diff
@@ -68,6 +68,8 @@ You can get a list of commands by typing the command ``\?``:
 +----------------------+------------------------------------------------------------------------------------+
 | ``\q``               | Quit Crate shell                                                                   |
 +----------------------+------------------------------------------------------------------------------------+
+| ``\check``           | Query the checks.sys table for failing cluster checks                              |
```
mfussenegger: `check.sys` ? It's `sys.checks` ?

PR #80, opened 2015-09-02
Use actual types returned by the crate python client

kovrus:


src/crate/crash/tabulate.py
```diff
@@ -538,7 +538,7 @@ def _column_type(strings, has_invisible=True):
     True
 
     """
-    types = [_type(s, has_invisible) for s in strings ]
```
dobe: strings as name implies that there are strings, which is according to your usage not the case

dobe:
add a test

dobe:
changes.txt

dobe:
ack

PR #83, opened 2015-09-21
Fixed bug in parsing internal non-sql commands

kovrus:


src/crate/crash/usage.txt
```diff
@@ -55,7 +55,7 @@ Successful connects will give you some information about the servers you connect
 If you connect to more than one server, the command will succeed
 if at least one server is reachable::
 
-    cr> \connect 127.0.0.1:44209 300.300.300.300:4295
+    cr> \connect 127.0.0.1:44209; 300.300.300.300:4295;
```
mfussenegger: imho this version shouldn't be valid.. semicolon within the line is an error

src/crate/crash/command.py
```diff
@@ -443,7 +443,9 @@ def _try_exec_cmd(self, line):
         words = line.split(' ', 1)
         if not words or not words[0]:
             return False
-        cmd = self.commands.get(words[0].lower())
+        cmd = self.commands.get(words[0].lower().strip(';'))
```
mfussenegger: use rstrip instead of strip. If the string starts with a semicolon it should be an error too

mfussenegger:
ack

PR #88, opened 2015-10-20
fix assertion error that could occur under python2.7

mfussenegger:


mfelsche:
ack if travis likes it

PR #89, opened 2015-10-21
updated crate-python to 0.13.6

chaudum:


CHANGES.txt
```diff
@@ -5,6 +5,13 @@ Changes for crash
 Unreleased
 ==========
 
+ - updated crate-python to ``0.13.6`` which includes following changes:
```
mfussenegger: imho no changelog entry needes, those changes are irrelevant for any users and we don't need to make a release for those either.

mfussenegger:
ack

PR #93, opened 2015-11-26
fixed python2 bug faulty csv output

celaus:


src/crate/crash/outputs.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import print_function
```
chaudum: still required?

src/crate/crash/outputs.py
```diff
@@ -122,10 +124,13 @@ def json(self, result):
         yield self.to_json_str(obj)
 
     def csv(self, result):
-        wr = csv.writer(self.writer, doublequote=False, escapechar='\\')
+        wr = csv.writer(self.writer, doublequote=False, escapechar='\\', quotechar="'")
         wr.writerow(result.cols)
-        for row in result.rows:
-            wr.writerow(row)
+
+        json_dumps = lambda r: json.dumps(r, sort_keys = True)
```
chaudum: in the changes you write it is only required for python 2.x
then you could do a version check here

celaus: For output consistency I think it makes sense to do this in both versions.

mikethebeer:
ack :santa:

PR #95, opened 2015-12-03
added json_row output format

mikethebeer:
prints a json dataset row-by-row in separate lines

### DIFF
--- a/src/crate/crash/outputs.py
+++ b/src/crate/crash/outputs.py
@@ -56,6 +56,7 @@ class OutputWriter(object):
             'raw': self.raw,
             'mixed': self.mixed,
             'dynamic': self.dynamic,
+            'json_row': self.json_row
         }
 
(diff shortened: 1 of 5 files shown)


src/crate/crash/outputs.py
```diff
@@ -144,6 +142,17 @@ def dynamic(self, result):
             return self.mixed(result)
         else:
             return self.tabular(result)
+    
+    def json_row(self, result):
+        obj = [dict(zip(result.cols, x)) for x in result.rows]
```
chaudum:
```py
rows = (dict(zip(result.cols, x)) for x in result.rows)
 for row in rows:
    ...
```

chaudum: yield inside loop!

chaudum:
ack. please wait for green travis

PR #97, opened 2015-12-13
add python 3.5 to travis.yml

mfussenegger:


.travis.yml
```diff
@@ -4,12 +4,20 @@ python:
     - 2.7
     - 3.3
     - 3.4
+    - 3.5
+    - pypy
+    - pypy3
+    - nightly
```
chaudum: would make nightly optional ;)

chaudum:
otherwise ack

.travis.yml
mfussenegger: done

PR #112, opened 2016-02-12
json output: order keys by the select list

mfussenegger:
E.g.

    select x, y

Results in

    {
        "x",
        "y"
    }

And

    select y, x

results in

    {
        "y",
        "x"
    }

Instead of having a random order


chaudum:
ack

PR #115, opened 2016-03-10
update prompt-toolkit to 0.59

mfussenegger:
Contains some bug fixes and performance improvements.


chaudum:
the versions of `Pygments` and `wcwidth` did not change, so would be nice if the position of them in the `versions.cfg` would not change.
otherwise ack

PR #116, opened 2016-03-10
move KeyboardInterrupt handling into repl

mfussenegger:
Catching KeyboardInterrupt in the CrateCmd makes it impossible to
interrupt processes that use the CrateCmd class (Like the doctests in
Crate do)

This commit moves the KeyboardInterrupt handling into the repl module so
that it will only be caught if crash is used interactively. (That was
the original intention when it was added.)


seut:
ack

PR #122, opened 2016-04-01
fix: cancel query statement on windows

mikethebeer:


src/crate/crash/printer.py
```diff
@@ -45,6 +45,8 @@ def write(self, line, end=''):
                 print(line.encode('utf-8').decode('ascii', 'replace'), end=end)
             except UnicodeEncodeError:
                 print(line.encode('utf-8').decode('ascii', 'ignore'), end=end)
+            except KeyboardInterrupt:
+                raise KeyboardInterrupt
```
bogensberger: just raise should also do it

mikethebeer:
fix is not necessary anymore. Bug was related to `prompt_toolkit` and is already updated in `master`.

PR #123, opened 2016-04-19
Enable/disable autocomplete

chaudum:
It is now possible to disable autocompletion using ``--no-autocomplete`` command line argument and to toggle autocomplete using the ``\autocomplete`` crash command.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,8 +5,12 @@ Changes for crash
 Unreleased
 ==========
 
+ - It is now possible to disable autocompletion using ``--no-autocomplete``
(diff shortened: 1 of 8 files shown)


mfussenegger:
I think we should consider adding a generic command for options that change the behaviour of crash e.g `\cset autocomplete on` or `\cset format dynamic`

But ack anyway

chaudum:
I think that is a good idea @mfussenegger , especially in combination with a configuration file.
Guess that is something for a fun Friday evening with a pint ... 🤔

ChrisChinchilla:
Some small fixes @chaudum

celaus:
ACK

PR #125, opened 2016-05-13
Update prompt-toolkit to 1.0.0

mfussenegger:
Includes a lot of of VI mode improvements!

And the following change makes our `get_height` override unnecessary:

 - Only reserve menu space when `complete_while_typing=True` or when
   there are completions to be displayed.


chaudum:
ack

PR #128, opened 2016-05-21
configuration file

chaudum:


mfussenegger:
lgtm :+1:

src/crate/crash/command.py
```diff
@@ -68,24 +69,65 @@ def emit(self, record):
 HISTORY_FILE_NAME = 'crash_history'
 HISTORY_PATH = os.path.join(USER_DATA_DIR, HISTORY_FILE_NAME)
 
+USER_CONFIG_DIR = user_config_dir("Crate", "Crate")
+CONFIG_FILE_NAME = 'crash.cfg'
+CONFIG_PATH = os.path.join(USER_CONFIG_DIR, CONFIG_FILE_NAME)
```
aelesbao: wouldn't it be easier for the user to use have default path as `$HOME/.crash`? it took me some googling to find out which would be the default path on mac os

bogensberger:
ack

src/crate/crash/command.py
mfussenegger: @aelesbao `$HOME/.crash` wouldn't work for windows users and putting everything just in the HOME directory has become kinda bad practice. There's the [XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) which we support so you can actually control where to put the config files by setting environment variables

PR #129, opened 2016-06-01
Remove note about DELETE rowcount

mfussenegger:
In Crate 0.55 the DELETE statements will have a rowcount (unless it's
done on a partitioned table and partitions are removed)


chaudum:
ack

PR #134, opened 2016-06-23
fixed read file command for legacy python

### LINKED ISSUE #133: \r <file> command does not work when using legacy Python 2
`open()` does not have `encoding` keyword` in python 2
https://github.com/crate/crash/blob/master/src/crate/crash/commands.py#L63

reported by @chdupuy in slack channel:
```
I only have this, even with crash -v : 
cr> \connect xxxxxxxxxxxxx:4200
+-------------------------------+------------------------------------------------------+---------+-----------+---------+
| server_url                    | node_name                                            | version | connected | message |
+-------------------------------+------------------------------------------------------+---------+-----------+---------+
| http://xxxxxxxxxxxxxxxxx:4200 |        crate-db-d393630c-ddd7-4cba-ab94-aa96889558ed | 0.54.9  | TRUE      | OK      |
+-------------------------------+------------------------------------------------------+---------+-----------+---------+
CONNECT OK
CLUSTER CHECK OK
cr> \r creation_base.sql
'encoding' is an invalid keyword argument for this function
help: r
read and execute statements from a file 
cr>
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35392789-r-file-command-does-not-work-when-using-python-2?utm_campaign=plugin&utm_content=tracker%2F505665&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F505665&utm_medium=issues&utm_source=github).
</bountysource-plugin>

chaudum:
fixes #133

### DIFF
--- a/src/crate/crash/commands.py
+++ b/src/crate/crash/commands.py
@@ -60,9 +60,9 @@ class ReadFileCommand(Command):
         return glob.glob(text + '*.sql')
 
     def __call__(self, cmd, filename, *args, **kwargs):
-        with open(filename, 'r', encoding='utf-8') as f:
+        with open(filename, 'rb') as f:
             for line in f:
-                cmd.process(line)
+                cmd.process(line.decode('utf-8'))
 
 
 class SwitchFormatCommand(Command):
--- a/src/crate/crash/test_commands.py
+++ b/src/crate/crash/test_commands.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8; -*-
+# vi: set encoding=utf-8
 # Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
 # license agreements.  See the NOTICE file distributed with this work for
 # additional information regarding copyright ownership.  Crate licenses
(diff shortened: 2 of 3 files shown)


src/crate/crash/test_commands.py
```diff
@@ -17,13 +19,18 @@
...
+import tempfile
 from unittest import TestCase
 from mock import patch
 from .commands import ReadFileCommand, ToggleAutocompleteCommand
 
 
 class ReadFileCommandTest(TestCase):
 
+    def setUp(self):
+        self.tmp_dir = tempfile.mkdtemp()
```
mfussenegger: > The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.

src/crate/crash/commands.py
```diff
@@ -60,9 +60,9 @@ def complete(self, cmd, text):
         return glob.glob(text + '*.sql')
 
     def __call__(self, cmd, filename, *args, **kwargs):
-        with open(filename, 'r', encoding='utf-8') as f:
+        with open(filename, 'rb') as f:
```
mfussenegger: fyi: Could also have used `os.open` - but this is just as good =)

src/crate/crash/test_commands.py
```diff
@@ -17,13 +19,23 @@
...
+
+    def tearDown(self):
+        if os.path.exists(self.tmp_dir):
+            shutil.rmtree(self.tmp_dir)
```
mfussenegger: fyi: There is also a `ignore_errors` parameter - has the advantage that there is no race condition

mfussenegger:
lgtm

PR #143, opened 2016-07-18
use correct statement for displaying node check in startup

kovrus:


src/crate/crash/command.py
```diff
@@ -275,7 +275,7 @@ def _connect(self, server):
             self.logger.info('CONNECT OK')
 
             """ check for failing node and cluster checks """
```
chaudum: there should not be a docstring
if, only a comment

chaudum:
lgtm

PR #144, opened 2016-07-22
added min requirement of urllib3 to be compatible

mikethebeer:


aslanbekirov:
ack

mikethebeer:
closed because fixed in https://github.com/crate/crate-python/pull/178

PR #146, opened 2016-07-27
rename the argument of the check command

kovrus:
`node` to `nodes`


src/crate/crash/commands.py
```diff
@@ -95,27 +95,28 @@ def __call__(self, cmd, *args, **kwargs):
 
 class CheckBaseCommand(Command):
 
-    CHECK_NAME = None
+    def check_name(self):
```
mfussenegger: could/should this be a property?

mfussenegger:
lgtm

PR #147, opened 2016-07-29
add hostname to output of `check nodes` command

kovrus:


joemoe:
wow, this is awesome.
ACK

PR #153, opened 2016-10-11
B/fix non et

bogensberger:


src/crate/crash/test_command.py
```diff
@@ -365,6 +366,23 @@ def test_error_exit_code(self):
             exception_code = e.code
         self.assertEqual(exception_code, 1)
 
+    def test_error_logging(self):
```
mfussenegger: Might make sense to split this up into two different tests with more specific method names

PR #154, opened 2016-11-07
crash was trying to connect to invalid hosts

chaudum:
if no hosts were provided via `--hosts` argument and configuration file was already present

### DIFF
--- a/src/crate/crash/command.py
+++ b/src/crate/crash/command.py
@@ -124,11 +124,11 @@ def get_parser(output_formats=[], conf=None):
     """
 
     def _conf_or_default(key, value):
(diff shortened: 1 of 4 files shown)


src/crate/crash/command.py
```diff
@@ -124,7 +124,7 @@ def get_parser(output_formats=[], conf=None):
     """
 
     def _conf_or_default(key, value):
-        return conf and conf.get_or_set(key, value) or value
+        return conf.get_or_set(key, value) if value is not None else value
```
bogensberger: ```return None if value is None else conf.get_or_set(key, value)``` is easier to read imho, but that's a matter of taste

PR #167, opened 2017-01-12
fix schema_name replacement

m-kharbat:


src/crate/crash/sysinfo.py
```diff
@@ -67,25 +66,12 @@ class SysInfoCommand(object):
...
-
-        self.CLUSTER_INFO.append(
-            information_schema_query.format(schema=schema_name))
 
     def execute(self):
         """ print system and cluster info """
         if not self.cmd.is_conn_avaliable():
             return
-        if self.cmd.connection.lowest_server_version >= SYSINFO_MIN_VERSION:
+        if self.cmd.connection.lowest_server_version >= SYSINFO_MIN_VERSION:
```
bogensberger: remove the extra spaces

src/crate/crash/command.py
```diff
@@ -285,6 +285,19 @@ def _connect(self, server):
...
+
+            information_schema_query = \
+                """ select count(distinct(table_name))
+                        as number_of_tables
+                    from information_schema.tables
+                    where {schema}
+                    not in ('information_schema', 'sys', 'pg_catalog') """
+
+            SysInfoCommand.CLUSTER_INFO.append(
+                information_schema_query.format(schema=schema_name))
```
bogensberger: please move that check to a method like
```
def information_schema_query(self, version):
....
```

and add a unit test for that method

src/crate/crash/command.py
```diff
@@ -285,6 +285,19 @@ def _connect(self, server):
...
+                TABLE_SCHEMA_MIN_VERSION else "schema_name"
+
+            information_schema_query = \
+                """ select count(distinct(table_name))
+                        as number_of_tables
+                    from information_schema.tables
+                    where {schema}
+                    not in ('information_schema', 'sys', 'pg_catalog') """
+
+            SysInfoCommand.CLUSTER_INFO.append(
```
chaudum: this will append every time you connect?

PR #169, opened 2017-01-24
auto-capitalize keywords while typing

chaudum:
crash-auto-capitalize

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ Changes for crash
 Unreleased
 ==========
 
+ - Automatically capitalize keywords while typing,
+   e.g. ``select`` -> ``SELECT``
+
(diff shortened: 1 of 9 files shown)


mikethebeer:
nice. However, I would like to have a setting that can be used to activate/deactivate this feature.

mikethebeer:
@chaudum noticed an issue that causes partly-capitalization of a word that is not registered as a keyword.

```sql
cr> SELECT SETtings FROM sys.cluster;
```

chaudum:
@mikethebeer addressed your comments
* added cli argument and runtime command: https://github.com/crate/crash/pull/169/commits/81853b71d30c1cd89a5cca9b3a513b913f6e831a
* improved capitalization: https://github.com/crate/crash/pull/169/commits/96d18f2fb85672f812bdd15c78d4ce97e30e9d56

/cc @Autophagy

PR #174, opened 2017-02-17
fixed autocapitalize cursor issue

m-kharbat:
- fixed the cursor issue http://g.recordit.co/I5kZVj8LQK.gif
- fixed when you type `select * from foo` then move the cursor back to `*` and add `date_trunc` it will capitalize DATE but not lowercase it again when typing further. http://g.recordit.co/hTB7BCuB0C.gif

### DIFF
--- a/src/crate/crash/repl.py
+++ b/src/crate/crash/repl.py
@@ -186,17 +186,17 @@ class Capitalizer:
         current_line = buffer.document.text
         cursor_position = buffer.document.cursor_position
 
-        if self.last_changed and self.is_prefix(current_line.lower(), self.last_changed.lower()):
+        if self.last_changed and self.is_prefix(current_line[:cursor_position].lower(), self.last_changed.lower()):
             diff = len(self.last_changed) - len(current_line)
             current_line = self.last_changed + current_line[diff:]
 
-        new_line = re.sub(self.KEYWORD_RE, self.keyword_replacer, current_line)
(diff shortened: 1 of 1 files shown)


src/crate/crash/repl.py
```diff
@@ -186,16 +186,19 @@ def __call__(self, buffer):
         current_line = buffer.document.text
         cursor_position = buffer.document.cursor_position
 
+
         if self.last_changed and self.is_prefix(current_line.lower(), self.last_changed.lower()):
             diff = len(self.last_changed) - len(current_line)
             current_line = self.last_changed + current_line[diff:]
 
-        new_line = re.sub(self.KEYWORD_RE, self.keyword_replacer, current_line)
+        new_line = re.sub(self.KEYWORD_RE, self.keyword_replacer, \
```
chaudum: `\` not required when having function arguments separated by newlines
same below

CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for crash
 Unreleased
 ==========
 
+ - Fixed issue that caused the autocapitalization to move the cursor
```
chaudum: we must not forget to remove that line when doing a release

src/crate/crash/repl.py
chaudum: also, I think all arguments would fit into a single line

CHANGES.txt
m-kharbat: should I remove it from the changelog ?

chaudum:
@m-kharbat please wait for green CI

PR #175, opened 2017-02-23
Stopped autocapitalisation of backslash preceded keywords

Autophagy:


src/crate/crash/repl.py
```diff
@@ -222,7 +222,7 @@ def is_multiline():
 
 class Capitalizer:
 
-    KEYWORD_RE = r'(?:"\w+)|(?:\'\w+)|\w+'
+    KEYWORD_RE = r'(?:"\w+)|(?:\'\w+)|(?:\\\w+)|\w+'
```
chaudum: this does not catch the case `\format dynamic`

You need to disable auto-captialisation when line starts with `\`

src/crate/crash/test_repl.py
```diff
@@ -83,6 +83,10 @@ def test_capitalize(self):
         self.capitalizer(buffer)
         self.assertEqual(u'CREATE TABLE test\n(a BOOLEAN, b STRING, c INTEGER)', buffer.text)
 
+        text = u'\select'
```
chaudum:
```py
text = u'\format dynamic'
```

src/crate/crash/repl.py
```diff
@@ -184,19 +233,21 @@ def __call__(self, buffer):
             return
 
         current_line = buffer.document.text
-        cursor_position = buffer.document.cursor_position
 
-        if self.last_changed and self.is_prefix(current_line[:cursor_position].lower(), self.last_changed.lower()):
-            diff = len(self.last_changed) - len(current_line)
-            current_line = self.last_changed + current_line[diff:]
+        if current_line[0] != '\\':
```
chaudum: code style: make as little indentation (nested if blocks) as possible

```py
if current_line.startswith('\\'):
    return
```

then you can leave the indentation of the rest as is

PR #177, opened 2017-02-24
make autocaptialization experimental and turn off by default

chaudum:


CHANGES.txt
```diff
@@ -5,11 +5,11 @@ Changes for crash
...
 ==========
 
- - Expanded the list of keywords used by autocompletion and 
+ - Expanded the list of keywords used by autocompletion and
    autocapitlization.
 
- - Automatically capitalize keywords while typing,
-   e.g. ``select`` -> ``SELECT``
+ - Automatically capitalize keywords while typing, e.g. ``select`` -> ``SELECT``
+   **This feature is experimental and may be removed in future versions!**
```
seut: shouldn't we mention that it's off by default? otherwise it implies to me that it is enabled from now on

chaudum: @seut ok, will also add a note in the changelog

chaudum:
Note that the default value for autocapitalize has changed from True to False. If you have already used this feature and you want to have autocapitalize turned off by default you also need to remove the default value from the crash config.

PR #185, opened 2017-03-29
improve PyPI docs

nslater:


nslater:
@chaudum I updated this to fix the same issue you mentioned on crate/crate-python#216. can you re-ack?

PR #189, opened 2017-06-26
Added support for SSL connection & username auth.

matriv:


src/crate/crash/command.py
```diff
@@ -472,6 +494,34 @@ def main():
     cmd.exit()
     sys.exit(cmd.exit_code)
 
+def test_main():
```
mfussenegger: please don't add test related code in this module

matriv: @mfussenegger any suggestion how to test then?

mfussenegger: Would it be possible to extract some smaller methods from `main()` - which are then suitable to be called from a test?

matriv: ok, makes sense, I'll try it

chaudum:
I guess the coverage decrease is a false positive.

src/crate/crash/command.py
```diff
@@ -472,6 +475,24 @@ def main():
     cmd.exit()
     sys.exit(cmd.exit_code)
 
+def createCmd(crate_hosts, error_trace, output_writer, is_tty, args):
```
mfussenegger: please use snake_case and prefix with `_`  (also in the tests)

matriv:
@chaudum the actual functionality of the params is tested in crate-python.
Also currently we don't have a CrateDB version released with SSL to write integration tests.
I added a task to do that once SSL functionality is released.

PR #190, opened 2017-06-27
Don't print the whole stack trace on Connection error.

matriv:
It now looks like that:
```
➭ bin/crash -v --verify-ssl true --cert-file /tmp/certs/marios.crt --key-file /tmp/certs/marios.key --ca-cert-file /tmp/certs/rootCA.crt --hosts https://127.0.0.1:4200 -U unknonwn
/Users/matriv/crate/crash/eggs/urllib3-1.14-py3.6.egg/urllib3/connection.py:266: SubjectAltNameWarning: Certificate for 127.0.0.1 has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)`
  SubjectAltNameWarning
401 Client Error: Unauthorized
```

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ Changes for crash
 Unreleased
 ==========
 
+ - Print only error message instead of full stacktrack in case of connection
+   error.
+
  - Added support for SSL connection to CrateDB.
 
  - Added new parameter ``username`` used to authenticate the user in CrateDB.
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+ - Removed printing of full stack trace in case of connection error.
```
chaudum: Print only error message instead of full stacktrack in case of connection error.

PR #195, opened 2017-06-30
Updated bootstrap.py for usage with a version parameter

christianbader:


versions.cfg
```diff
@@ -49,7 +49,7 @@ colorama = 0.3.6
 # zope.exceptions==4.0.8
 # zope.interface==4.1.3
 # zope.testrunner==4.4.10
-setuptools = 34.3.3
+# setuptools = 34.3.3
```
mfussenegger: why is this now commented out?

christianbader: because this causes an VersionConflict with setuptools 34.1.1

mfussenegger: Shouldn't we update the usage instructions to create a virtualenv that dosn't contain a setuptools copy? Otherwise every time we update the versions.cfg we've to take care to re-disable setuptools pinning.

PR #197, opened 2017-07-03
Validate ssl related filepaths beforehand.

matriv:


src/crate/crash/command.py
```diff
@@ -481,11 +481,22 @@ def main():
     sys.exit(cmd.exit_code)
 
 def _create_cmd(crate_hosts, error_trace, output_writer, is_tty, args):
+
+    # Validate paths for certificate and key files
+    cert_file = args.cert_file
```
mfussenegger: Did you consider using the `type` functionality of argparse?
E.g. something like this:

```
def file_with_permissions(f):
    if not os.access(f, os.R_OK):
        raise Exception("foo")
    return f


parser = argparse.ArgumentParser()
parser.add_argument('--file', type=file_with_permissions, required=True)
try:
    parser.parse_args()
except Exception as e:
    print(str(e))
```

matriv: no didn't know that :-) thx!

src/crate/crash/command.py
```diff
@@ -499,5 +503,10 @@ def _create_cmd(crate_hosts, error_trace, output_writer, is_tty, args):
                     ca_cert_file=args.ca_cert_file,
                     username=args.username)
 
+def file_with_permissions(path):
+    if not os.access(path, os.R_OK):
+        raise Exception('[' + path + '] does not exist or is not readable')
```
mfussenegger: I think it would make sense to raise a exception that is a bit more specific here

matriv: What kind of Exception? Should I create a new class like ValidationError or ParseError?

PR #202, opened 2017-07-04
Use python3.6 to build crash_standalone.

matriv:
Otherwise we end up with this error:
TypeError: <crate.crash.repl.Capitalizer instance at 0x108b52908> is not a Python function


matriv:
There is a fix to make it compatible with python2.7 so this is not necessary

PR #212, opened 2017-07-21
Add support for tab indentation & backspace de-indentation

mfussenegger:
Adds support for indentation via tab. Only works in multi-line:

The following does nothing:

```
<tab>
```

This works:

```
create table t1 x (
<tab> -> 4 spaces
```

De-indentation:
```
create table t1 x(
    <bs> -> removes 4 spaces
```

In addition the auto-completion popup is only shown if at least 1 character has been typed.

### DIFF
--- a/src/crate/crash/repl.py
+++ b/src/crate/crash/repl.py
@@ -46,6 +46,7 @@ from prompt_toolkit.layout.processors import (
     ConditionalProcessor
 )
 from prompt_toolkit.key_binding.manager import KeyBindingManager
+from prompt_toolkit.keys import Keys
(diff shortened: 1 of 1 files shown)


src/crate/crash/repl.py
```diff
@@ -194,6 +195,9 @@ def get_completions(self, document, complete_event):
             return
         line = document.text
         word_before_cursor = document.get_word_before_cursor()
+        # Show completion only if at least 1 character has been typed
```
mikethebeer: autocompletion on 3rd character is already in master. https://github.com/crate/crash/commit/6a6070a14ce86a83cdddc8b7d60cd2cb225e8016  --> pls rebase.

PR #217, opened 2017-08-09
added session toolbar

mikethebeer:
* shows the current session user
* shows the handler nodes connected with

crash-toolbar

### DIFF
--- a/src/crate/crash/layout.py
+++ b/src/crate/crash/layout.py
@@ -1,0 +1,156 @@
+# vim: set fileencodings=utf-8
+# -*- coding: utf-8; -*-
+# PYTHON_ARGCOMPLETE_OK
+#
+# Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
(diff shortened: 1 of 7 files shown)


src/crate/crash/layout.py
```diff
@@ -0,0 +1,123 @@
...
+from prompt_toolkit.layout.processors import ConditionalProcessor, HighlightSearchProcessor
+from prompt_toolkit.layout.toolbars import TokenListToolbar, SearchToolbar
+from prompt_toolkit.layout.margins import PromptMargin, ConditionalMargin
+from prompt_toolkit.layout.utils import token_list_width
+
+__all__ = (
+    'create_layout',
+)
+
+def create_layout(message='', lexer=None,
```
mfussenegger: please add a comment explaining the difference to `create_prompt_layout`

src/crate/crash/layout.py
```diff
@@ -0,0 +1,123 @@
...
+from prompt_toolkit.layout.controls import TokenListControl, BufferControl, FillControl
+from prompt_toolkit.layout.lexers import PygmentsLexer
+from prompt_toolkit.layout.screen import Char
+from prompt_toolkit.layout.menus import CompletionsMenu
+from prompt_toolkit.layout.processors import ConditionalProcessor, HighlightSearchProcessor
+from prompt_toolkit.layout.toolbars import TokenListToolbar, SearchToolbar
+from prompt_toolkit.layout.margins import PromptMargin, ConditionalMargin
+from prompt_toolkit.layout.utils import token_list_width
+
+__all__ = (
```
mfussenegger: no need to define `__all__` as we don't do a `import *`

src/crate/crash/repl.py
```diff
@@ -274,21 +275,40 @@ def create_buffer(cmd, history_file):
 
 
 def loop(cmd, history_file):
+
+    def session_toolbar(cli):
+        tokens = []
+        if cmd.is_conn_avaliable():
+            active_servers = cmd.connection.client.active_servers
+            active_servers = [n.replace('http://', '') for n in active_servers]
+            active_servers = [n.replace('https://', '') for n in active_servers]
```
mfussenegger: could use generator expressions instead of list comprehensions to avoid the copying

PR #218, opened 2017-08-17
added support for multiline table headers in the tabular output format

Taliik:
```
cr> select 1 = (select 2);
+----------------+
| (1 = (SELECT 2 |
|                |
| ))             |
+----------------+
| FALSE          |
+----------------+
SELECT 1 row in set (0.003 sec)
cr>
```

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+ - added support for multiline table headers in the tabular output format
+ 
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+ - added multiline support
```
mikethebeer: I would write it like this:
> added support for multiline table headers in the tabular output format.

mikethebeer:
pls squash before merge

PR #222, opened 2017-08-24
crash multiline header fix

Taliik:
fixed the problem with the formatting if the first row was empty, in line 585 & 586
also added some formatting tests

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+ - Added back support for multiline table headers
+
 2017/08/21 0.22.1
 =================
 
--- a/src/crate/crash/tabulate.py
+++ b/src/crate/crash/tabulate.py
@@ -30,7 +30,6 @@ import re
 import sys
 
 
-
 if sys.version_info[:2] == (2, 7):
     def float_format(val):
         return repr(val)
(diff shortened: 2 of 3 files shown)


src/crate/crash/tabulate.py
```diff
@@ -506,24 +547,61 @@ def _align_column(strings, alignment, minwidth=0, has_invisible=True):
...
+                "\n".join([padfn(maxwidth, s) for s in ms.splitlines()])
+                for ms in strings]
+        else:
+            # enable wide-character width corrections
+            s_lens = [max((len(s) for s in re.split("[\r\n]", ms))) for ms in strings]
+            visible_widths = [maxwidth - (w - l) for w, l in zip(s_widths, s_lens)]
+            # wcswidth and _visible_width don't count invisible characters;
+            # padfn doesn't need to apply another correction
+            if strings[0] == '':
+                strings[0] = ' '
```
mxm: This seems a bit hacky. Can't we enable support for empty string padding in `padfn`?

mxm: Are these the only changes from the old multiline formatting code?

Taliik: Yes it kinda is. I don't think padfn supports empty strings, but i will have another look to see if i can find a better solution.
yes these are the only changes.

CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+ - added back support for multiline table headers
```
mxm: You could capitalize the first letter here.

PR #238, opened 2017-11-03
Don't execute lines containing only whitespaces

mfussenegger:
A file with trailing whitespace could lead to something like this:

    CONNECT OK
    DROP OK, 1 row affected  (0.058 sec)
    CREATE OK, 1 row affected  (0.193 sec)
    SQLActionException[Failed to parse source [{"stmt": ""}]]


mfussenegger:
> changelog entry?

I think that code was still unreleased and in the released version it appears to have been ignored already

PR #240, opened 2017-11-07
Fixed CrateCmd having a missing last_connected_servers attribute on initialisation

Autophagy:


CHANGES.txt
```diff
@@ -23,6 +23,9 @@ Fixes
    crash in verbose mode. This prevents the programme from printing the message
    to stdout when using the command mode or executing ``--sysinfo``.
 
+ - Fixed an error caused by using the `\verbose` command before using `\connect`
```
chaudum: use double backticks for inline code
and please move on top of list

Autophagy: Aha, thanks. Been a while since I did a changelog entry...

PR #243, opened 2017-12-03
Update buildout version

mfussenegger:
Jenkins started failing. For some reason it picked up buildout 2.9.6 and
then failed because it doesn't match the required version 2.9.5


mfussenegger:
Was updated to 2.10 in another PR

PR #245, opened 2017-12-13
added support for HTTP password authentication

mikethebeer:
If HTTP password authentication is configured on the server, Crash
prompts for a password when connect.

crash-pw-auth

Introduced new cli parameter `-W` that force prompts for password immediately to avoid an unnecessary try-connect-request against the server.

**NOTE**:
`is_tty` check and the `getpass()` prompt makes it very difficult to unit test. Therefore the automatic password prompt is only tested manually.


src/crate/crash/command.py
```diff
@@ -310,13 +316,21 @@ def _do_connect(self):
...
 
     def _connect(self, server):
         """ connect to the given server, e.g.: \connect localhost:4200 """
         self.last_connected_servers = server
-        self._do_connect()
+        try:
+            self._do_connect()
+        except (ProgrammingError, LocationParseError) as e:
+            if ('401 Client Error: Unauthorized' in e.args[0]):
+                self.password = getpass()
```
matriv: In my opinion it would be nice to also prompt for username here.

matriv:
@mikethebeer I think that we need also a way to provide a password if we execute an sql stmt with `crash -U -c <stmt>`. psql uses the `~/.pgpass` but also an env variable `PGPASSWORD`.  Check: https://stackoverflow.com/questions/6405127/how-do-i-specify-a-password-to-psql-non-interactively/6405162
If we don't have something like that then a user which is authenticated with password cannot execute crash -c ...

@mfussenegger  maybe you have another suggestion?

mxm:
@matriv @mikethebeer An environment variable like `CRATEPASSWORD` seems the best choice. I think we can do without the password file for now.

Just curious, why do we really need the new `-W` flag? It seems like a tiny optimization which comes at the price of maintaining more code. I think we could live without this parameter.

matriv:
@mxm I'm positive to keep the `-W` as this keeps consistency with psql behaviour

src/crate/crash/command.py
```diff
@@ -494,9 +509,42 @@ def main():
...
     crate_hosts = [host_and_port(h) for h in args.hosts]
     error_trace = args.verbose > 0
+
+    force_passwd_prompt = args.password
+    username = args.username
+    password = None
+
+    # If no user provided check env. variables LOGNAME, USER, LNAME and USERNAME
+    if not username:
+        username = getuser()
```
mfussenegger: If we support this implicit user retrieval I think we also have to document it. Could also be changed to `username = args.username or getuser()` in the definition further above.

src/crate/crash/command.py
```diff
@@ -494,9 +509,42 @@ def main():
...
+
+    # If password prompt is not forced try to get it from env. variable.
+    if not force_passwd_prompt:
+        password = os.environ.get('PGPASSWORD', '')
```
mfussenegger: Same here, I think we have to document this.

src/crate/crash/command.py
```diff
@@ -494,9 +509,42 @@ def main():
...
+
+    # Prompt for password immediately to avoid that the first time trying to
+    # connect to the server runs into an `Unauthorized` excpetion
+    # is_tty = False
+    if force_passwd_prompt and is_tty:
```
mfussenegger: this means that even if the env variable is set, we still prompt?

src/crate/crash/command.py
```diff
@@ -310,13 +316,22 @@ def _do_connect(self):
...
         self.cursor = self.connection.cursor()
 
     def _connect(self, server):
         """ connect to the given server, e.g.: \connect localhost:4200 """
         self.last_connected_servers = server
-        self._do_connect()
+        try:
+            self._do_connect()
+        except (ProgrammingError, LocationParseError) as e:
+            if ('401 Client Error: Unauthorized' in e.args[0]):
```
mfussenegger: I think it would be nice to have a specific exception that is raised for Unauthorized errors - so that we don't have to parse the error in multiple places. (Ideally we would've done that in crate-python, but we can also do it here, by wrapping the connect method or encapsulate it in CrateCmd somehow)

src/crate/crash/command.py
```diff
@@ -310,13 +316,22 @@ def _do_connect(self):
...
+                self.username = input('Username: ')
+                self.password = getpass()
```
mfussenegger: as discussed: From a separation point of view it would be nice if the `CrateCmd` didn't do any user-input handling. That should be up to the `repl` (or `main` to establish the initial connection)

src/crate/crash/command.py
mikethebeer: good catch. no, in this case we shouldn't prompt.

CHANGES.txt
```diff
@@ -8,6 +8,8 @@ Unreleased
 Changes
 -------
 
+ - Added support for HTTP password authentication.
```
chaudum: I would only say: Added support for password authentication.
HTTP is implementation detail, maybe users are not even aware that crate uses HTTP protocol.

docs/cli.txt
```diff
@@ -79,6 +82,25 @@ Example Usage with SSL
...
+=====================
+Environment Variables
+=====================
+
+Crash takes the following environment variables into account to select default
+connection parameters. These can be helpful to avoid user-input prompts or any
+other hard-coded cli parameter.
+
+:CRATEPWD: Password to be used if CrateDB demands password authentication.
+           For security reasons the usage of this variable is not recommended.
```
mxm: We introduce a new feature but we discourage people from using it? I don't think that is an issue if the environment variable is not exposed.

docs/cli.txt
```diff
@@ -15,6 +15,9 @@ The ``crash`` binary supports several command line arguments.
 +---------------------------------------+--------------------------------------------------+
 | ``-U``, ``--username``                | The username to authenticate in the database     |
 +---------------------------------------+--------------------------------------------------+
+| ``-W``, ``--password``                | Force prompt for password. Otherwise it prompts  |
+|                                       | for password automatically.                      |
```
chaudum: in case the hba configuration of the server requires "password authentication" for the specified user ?

docs/cli.txt
```diff
@@ -79,6 +82,25 @@ Example Usage with SSL
...
+
+:CRATEPWD: Password to be used if CrateDB demands password authentication.
+           For security reasons the usage of this variable is not recommended.
+
+
+For example::
+
+    $ export CRASHPWD=dustins-password
+    $ crash --hosts localhost:4200 \
+            --username dustin
```
chaudum: missing \

docs/cli.txt
```diff
@@ -79,6 +82,25 @@ Example Usage with SSL
...
+            -c 'select * from sys.nodes' > output.json
```
chaudum: default format is "tabular" but your command suggests json output

src/crate/crash/command.py
```diff
@@ -360,6 +367,8 @@ def _try_exec_cmd(self, line):
                     message = cmd(self, *words[1:])
                 else:
                     message = cmd(*words[1:])
+            except ProgrammingError as e:
+                raise e
```
chaudum: print error rather than raising exception?

src/crate/crash/command.py
```diff
@@ -494,9 +503,37 @@ def main():
 
     crate_hosts = [host_and_port(h) for h in args.hosts]
     error_trace = args.verbose > 0
+
+    force_passwd_prompt = args.password
+    password = None
+
+    # If password prompt is not forced try to get it from env. variable.
+    if not force_passwd_prompt:
+        password = os.environ.get('CRATEPWD', '')
```
chaudum: password should stay `None` if CRATEPWD is not available

chaudum: in my opinion, PWD is not a common abbreviation for password.
either PW or PASSWD

src/crate/crash/command.py
```diff
@@ -494,9 +503,37 @@ def main():
...
+        password = getpass()
+
+    # Tries to create a connection to the server.
+    # Prompts for the password automatically if the server only accepts
+    # password authentication.
+    cmd = None
     try:
-        cmd = _create_cmd(crate_hosts, error_trace, output_writer, is_tty, args)
-    except (ProgrammingError, LocationParseError) as e:
+        try:
```
chaudum: it's not necessary to wrap it into another try/except block. you can write:

```py
try:
    ...
except (ProgrammingError, LocationParseError) as e:
   ...
except Exception as e:
   ...
```

src/crate/crash/command.py
```diff
@@ -494,9 +503,37 @@ def main():
...
+            cmd = _create_cmd(crate_hosts, error_trace, output_writer, is_tty,
+                              args, password=password)
+        except (ProgrammingError, LocationParseError) as e:
+            if ('401' in e.message) and not force_passwd_prompt:
```
chaudum: minor: parenthesis are not necessary

docs/cli.txt
chaudum: Agree, however we should also state that storing passwords in an env variable is not safe ;)

src/crate/crash/command.py
mikethebeer: this is necessary because otherwise the REPL (which calls the user-input prompt) won't receive any exception.

mikethebeer:
@chaudum @mfussenegger pushed some improvements. Can you take a 👀  again?

src/crate/crash/command.py
chaudum: got it

src/crate/crash/command.py
```diff
@@ -29,6 +29,7 @@
 import re
 import sys
 import urllib3
+from getpass import getpass, getuser
```
chaudum: getuser is unused

src/crate/crash/command.py
```diff
@@ -494,9 +503,40 @@ def main():
...
-        cmd = _create_cmd(crate_hosts, error_trace, output_writer, is_tty, args)
+        cmd = _create_cmd(crate_hosts, error_trace, output_writer, is_tty,
+                            args, password=password)
     except (ProgrammingError, LocationParseError) as e:
+        if '401' in e.message and not force_passwd_prompt:
+            if is_tty:
+                password = getpass()
+            try:
+                cmd = _create_cmd(crate_hosts, error_trace, output_writer,
+                                is_tty, args, password=password)
```
chaudum: indentation: is_tty should be under crate_hosts

src/crate/crash/command.py
```diff
@@ -494,9 +503,40 @@ def main():
...
+        password = getpass()
+
+    # Tries to create a connection to the server.
+    # Prompts for the password automatically if the server only accepts
+    # password authentication.
+    cmd = None
     try:
-        cmd = _create_cmd(crate_hosts, error_trace, output_writer, is_tty, args)
+        cmd = _create_cmd(crate_hosts, error_trace, output_writer, is_tty,
+                            args, password=password)
```
chaudum: indentation: args should be under crate_hosts

src/crate/crash/command.py
mfussenegger: Could add a comment along the lines of `# repl needs to handle authorization errors`

src/crate/crash/command.py
```diff
@@ -136,6 +137,9 @@ def _conf_or_default(key, value):
                         help='use -a to enable experimental auto-capitalization of SQL keywords')
     parser.add_argument('-U', '--username', type=str,
                         help='the username to authenticate in the database')
+    parser.add_argument('-W', '--password', action='store_true',
+                        dest='password', default=_conf_or_default('force_passwd_prompt', False),
```
mfussenegger: would it make sense to rename the `dest` parameter here already to `force_prompt_pwd` ? Reading the code further below I first thought this already contains the password

src/crate/crash/command.py
```diff
@@ -494,9 +503,40 @@ def main():
...
+    if not force_passwd_prompt:
+        password = os.environ.get('CRATEPW', None)
+
+    # Prompt for password immediately to avoid that the first time trying to
+    # connect to the server runs into an `Unauthorized` excpetion
+    # is_tty = False
+    if force_passwd_prompt and not password and is_tty:
+        password = getpass()
+
+    # Tries to create a connection to the server.
```
mfussenegger: I think this is pretty self explanatory and the comment wouldn't be necessary, but feel free to keep it.

PR #251, opened 2018-02-03
enabled sitemap generation through sphinx module in the docs

joemoe:
* added extension to `docs/conf.py`
* set base url in `docs/conf.py`
* this will automatically generate a sitemap for every docs build

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,1 +1,4 @@
 from crate.theme.rtd.conf.crash import *
+
+site_url = 'https://crate.io/docs/clients/crash/en/latest/'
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -1 +1,4 @@
 from crate.theme.rtd.conf.crash import *
+
+site_url = 'https://crate.io/docs/clients/crash/en/latest/'
+extensions = ['sphinx_sitemap']
```
jodok: we usually put a new line at the end of the files.
in sublime text you could add this to your settings: `	"ensure_newline_at_eof_on_save": true,
`
otherwise ack

PR #253, opened 2018-03-25
clean virtualenv on travis to avoid dependency problems

chaudum:
Please merge before #252

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,11 @@ python:
     - nightly
 
 before_install:
-    - pip install --upgrade setuptools
-    - pip uninstall -y six
+  # clean existing dependencies to avoid clashes
+  - pip freeze | xargs pip uninstall -y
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,8 +8,11 @@ python:
     - nightly
 
 before_install:
-    - pip install --upgrade setuptools
-    - pip uninstall -y six
+  # clean existing dependencies to avoid clashes
+  - pip freeze | xargs pip uninstall -y
+  - pip uninstall -y setuptools
+  - pip install setuptools==33.1.1  # pin specific version in virtualenv
```
mxm: Why install? Wouldn't it work to only uninstall? Buildout should install a new version.

chaudum: @mxm No it does not. It would fail with `ModuleNotFoundError: No module named 'pkg_resources'` in Python 3.6 or newer

PR #254, opened 2018-03-27
Updated \dt command to only show tables

chaudum:


chaudum:
@mfussenegger there is no user facing change. so I omitted the changelog entry

PR #257, opened 2018-04-12
rename CrateCmd to CrateShell

chaudum:
and other cleanups

### DIFF
--- a/src/crate/crash/command.py
+++ b/src/crate/crash/command.py
@@ -208,7 +208,7 @@ def _parse_statements(lines):
         yield ' '.join(parts)
 
 
-class CrateCmd(object):
+class CrateShell:
 
     def __init__(self,
                  output_writer=None,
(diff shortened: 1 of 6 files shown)


src/crate/crash/command.py
```diff
@@ -240,15 +248,19 @@ def __init__(self,
         }
         self.commands.update(built_in_commands)
         self.logger = ColorPrinter(is_tty)
+
+        self.output_writer = output_writer or \
+            OutputWriter(PrintWrapper(), is_tty)
```
mfussenegger: I think we should add a `setup.cfg` `pycodestyle` entry and raise the line-length to something like 88  (while still aiming for 80 in most cases)

Breaks like this imho hurt readability more than they improve it. See also https://github.com/ambv/black#line-length

chaudum: It's better than it was before, though, I think. but I agree, that it makes sense to put it into a single line for readability.

PR #258, opened 2018-04-20
docs rewrite take 2

nslater:


docs/run.txt
```diff
@@ -130,8 +135,8 @@ For example::
 
 .. _commands:
 
-==============
-Crash Commands
+
+Commands
 ==============
```
chaudum: does the title underline need to be the same length as the title itself?

PR #262, opened 2018-04-30
add flake8 linter script and fix resulting issues

chaudum:


buildout.cfg
```diff
@@ -32,6 +33,17 @@ working-directory = ${buildout:directory}/docs
 defaults = ['--auto-color']
 eggs = crash [test]
 
+[linter]
+recipe = zc.recipe.egg:script
+eggs = flake8
+relative-paths=true
+scripts = flake8=linter
+initialization =
+    sys.argv.extend(['--ignore', 'E501,C901',
```
mfussenegger: can we add these options in a `setup.cfg` file instead so that they're also picked up by linter editor-plugins?

chaudum: ah, did not know that this was possible. sure thing, will change it.

chaudum:
@mfussenegger updated.
@matriv can you have a look?

PR #263, opened 2018-05-03
add default schema support

chaudum:
It is now possible to use a default schema that is used when statements do not provide a schema explicitly. The default schema can be specified using the `--schema` startup parameter.


src/crate/crash/command.py
```diff
@@ -368,6 +378,22 @@ def _verify_connection(self, verbose=False):
                 # check for failing node and cluster checks
                 built_in_commands['check'](self, startup=True)
 
+    def _fetch_session_info(self):
+        if self.is_conn_available() \
+                and self.connection.lowest_server_version >= StrictVersion("2.0"):
+            self.cursor.execute("""
+                SELECT settings['license']['enterprise'] FROM sys.cluster;
+            """)
+            user_function = "current_user" if self.cursor.fetchone()[0] else "NULL"
```
mfussenegger: maybe add a short command that `current_user` is only available if enterprise is enabled.

src/crate/crash/command.py
```diff
@@ -141,6 +143,8 @@ def _conf_or_default(key, value):
     parser.add_argument('-W', '--password', action='store_true',
                         dest='force_passwd_prompt', default=_conf_or_default('force_passwd_prompt', False),
                         help='force a password prompt')
+    parser.add_argument('-s', '--schema', type=str,
```
mfussenegger: Would it make sense to only allow the long option name? In case we want to use `-s` for something else, e.g. for pgsql compat or something.

chaudum: Good point!

PR #264, opened 2018-05-04
clean up test code to get rid of resource warnings

chaudum:


src/crate/crash/test_command.py
```diff
@@ -21,6 +22,23 @@
...
+def crate_shell(hosts='localhost:4200', **kwargs):
+    """
+    Create an "auto-closable" CrateShell object that can be used in conjunction
+    with the ``with`` statement, for example::
+
+    >>> with crate_shell(error_trace=True) as cmd:
+    ...     # do something here
+
+    """
+    cmd = CrateShell(hosts.split(' '), **kwargs)
```
chaudum: I know this could be achieved by adding `__enter__` and `__exit__` methods to the CrateShell class, but I am reluctant to do so, as this is only for suppressing the warning in the test suite.

src/crate/crash/test_command.py
```diff
@@ -21,6 +22,23 @@
 from distutils.version import StrictVersion
 
 
+@contextmanager
```
chaudum: I know this could be achieved by adding `__enter__` and `__exit__` methods to the CrateShell class, but I am reluctant to do so, as this is only for suppressing the warning in the test suite.

mfussenegger: What's speaking against supporting the contextmanager generally?

src/crate/crash/command.py
```diff
@@ -363,6 +364,10 @@ def _verify_connection(self, verbose=False):
             else:
                 results.append(infos + (True, 'OK', ))
 
+        # sort by CONNECTED DESC, SERVER_URL
+        results.sort(key=itemgetter(3), reverse=True)
+        results.sort(key=itemgetter(0))
```
mfussenegger: doesn't the second sort overwrite the first?

mfussenegger: Okay, looked like depending on an implementation detail. But the python docs also state it explicitly.

src/crate/crash/test_command.py
chaudum: As I said, it's mainly for testing. However, I don't have a strong opinion.

mfussenegger: I think it could generally be useful, so I'd be in favour of adding it to the `CrateShell` class.

PR #274, opened 2018-06-28
Preserve newlines to be able to support line comments properly

mxm:
Removing newlines has some unwanted side effects like line comments not
terminating at the end of the line anymore.


mxm:
CC @emilywoods

PR #277, opened 2018-09-13
update install docs

nslater:
for https://trello.com/c/7bdvVTnH/3505

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,36 +27,16 @@ The CrashDB Shell (aka *Crash*) is an interactive `command-line interface`_
 Screenshot
 ==========
 
-.. image:: docs/query.png
+.. image:: https://raw.githubusercontent.com/crate/crash/master/docs/query.png
     :alt: A screenshot of Crash
-
-Prerequisites
-=============
-
-Recent versions of Crash require Python 3 (>= 3.4) to run.
-
-Use Crash version 0.23.x if you are running Python 2.7 or 3.3.
-
-Use Crash version 0.16.x if you're running Python 2.6.
 
 Installation
 ============
 
-Under normal circumstances, you do not have to manually install Crash.
-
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -27,93 +27,9 @@ The CrashDB Shell (aka *Crash*) is an interactive `command-line interface`_
 Screenshot
 ==========
 
-.. image:: docs/query.png
+.. image:: https://raw.githubusercontent.com/crate/crash/master/docs/query.png
     :alt: A screenshot of Crash
 
-Prerequisites
```
chaudum: I think the README must still contain installation instructions.

README.rst
```diff
@@ -27,93 +27,9 @@ The CrashDB Shell (aka *Crash*) is an interactive `command-line interface`_
...
-Recent versions of Crash require Python 3 (>= 3.4) to run.
-
-Use Crash version 0.23.x if you are running Python 2.7 or 3.3.
-
-Use Crash version 0.16.x if you're running Python 2.6.
-
-Installation
-============
-
-Under normal circumstances, you do not have to manually install Crash.
```
chaudum: Only line 45-54 are obsolete now.

README.rst
nslater: I do not want to duplicate the installation instructions. it is bound to happen that one of them is updated and the other one is missed

so in choosing between the README docs and the official website docs, I chose the official website docs as the preferred location

if we want to include installation help in the README, I would suggest that we add a new section pointing to the official docs for further help with installing

is that acceptable?

chaudum: I think this would be a bad user experience for developers. They look at the Github readme for instructions, not on the website. The instructions in the readme and on the website can differ, though. For the readme, the `pip install` command the the `curl` command to download the standalone version are totally sufficient.

PR #286, opened 2019-06-07
Upgrade prompt toolkit to 2.0.x

mfussenegger:


setup.py
```diff
@@ -30,7 +30,7 @@
     'Pygments>=2.0',
     'crate>=0.22.0',
     'appdirs>=1.2,<2.0',
-    'prompt-toolkit>=1.0,<1.1'
+    'prompt-toolkit>=2.0'
```
chaudum: Do you think it's save to allow everything `>=` 2.0?
The `versions.cfg` specifies `2.0.9`, so we don't test with greater versions.

mfussenegger: Added a `<3.0` constraint (dev branch is on 3.0, so it is unlikely that there are going to be 2.1 releases and if so I think they won't contain breaking changes.)

PR #289, opened 2019-07-05
Make --verify-ssl choices explicit

mfussenegger:
This adds true/false as explicit choices so that the help output is a
bit friendlier.

    [--verify-ssl {True,False}]
    --verify-ssl {True,False}
          Enable or disable the verification of the server SSL
          certificate

In CLIs with boolean flags it is also sometimes the case that you've a
``--no-*`` variant or that you can use ``no`` or something like that. So
having to pass ``false`` is not necessarily intuitive.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,6 @@ universal=1
 [flake8]
 ignore = E501,C901
 exclude = tabulate.py
+
+[pycodestyle]
+ignore = E501
--- a/src/crate/crash/command.py
+++ b/src/crate/crash/command.py
@@ -31,7 +31,7 @@ import sys
 import urllib3
 from getpass import getpass
 from appdirs import user_data_dir, user_config_dir
-from argparse import ArgumentParser
+from argparse import ArgumentParser, ArgumentTypeError
 from collections import namedtuple
 from crate.client import connect
 from crate.client.exceptions import ConnectionError, ProgrammingError
@@ -110,12 +110,14 @@ def parse_args(parser):
 
 
 def boolean(v):
-    if str(v).lower() in ("yes", "true", "t", "1"):
+    v = str(v).lower()
+    if v in ("yes", "true", "t", "1"):
         return True
-    elif str(v).lower() in ("no", "false", "f", "0"):
+    elif v in ("no", "false", "f", "0"):
         return False
     else:
-        raise ValueError('not a boolean value')
+        raise ArgumentTypeError(
+            'Invalid choice `{v}`, expected one of: [yes, true, t, 1, no, false, f, 0]'.format(v=v))
 
 
 def get_parser(output_formats=[], conf=None):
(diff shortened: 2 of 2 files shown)


mfussenegger:
Ah right, should we list all choices then? Or maybe go with `('true', 'false', 'yes', 'no')` ?

mfussenegger:
@chaudum pushed a fixup so that the error message is also friendlier if an invalid value is used:

```
crash --verify-ssl foobar
usage: crash [-h] [-v] [-A] [-a] [-U USERNAME] [-W] [--schema SCHEMA]
             [--history FILENAME] [--config FILENAME]
             [-c STATEMENT | --sysinfo] [--hosts [HOSTS [HOSTS ...]]]
             [--verify-ssl {true,false,yes,no}] [--cert-file FILENAME]
             [--key-file FILENAME] [--ca-cert-file FILENAME] [--format FORMAT]
             [--version]
crash: error: argument --verify-ssl: Input `foobar` cannot be converted to a boolean, expected one of: [yes, true, t, 1, no, false, f, 0]
```

chaudum:
> @chaudum pushed a fixup so that the error message is also friendlier if an invalid value is used:
> 
> ```
> crash --verify-ssl foobar
> usage: crash [-h] [-v] [-A] [-a] [-U USERNAME] [-W] [--schema SCHEMA]
>              [--history FILENAME] [--config FILENAME]
>              [-c STATEMENT | --sysinfo] [--hosts [HOSTS [HOSTS ...]]]
>              [--verify-ssl {true,false,yes,no}] [--cert-file FILENAME]
>              [--key-file FILENAME] [--ca-cert-file FILENAME] [--format FORMAT]
>              [--version]
> crash: error: argument --verify-ssl: Input `foobar` cannot be converted to a boolean, expected one of: [yes, true, t, 1, no, false, f, 0]
> ```

Afaik the choices must the possible return values of the `type` function:

```console
$ crash --verify-ssl true
usage: crash [-h] [-v] [-A] [-a] [-U USERNAME] [-W] [--schema SCHEMA]
             [--history FILENAME] [--config FILENAME]
             [-c STATEMENT | --sysinfo] [--hosts [HOSTS [HOSTS ...]]]
             [--verify-ssl {true,false,yes,no}] [--cert-file FILENAME]
             [--key-file FILENAME] [--ca-cert-file FILENAME] [--format FORMAT]
             [--version]
crash: error: argument --verify-ssl: invalid choice: True (choose from 'true', 'false', 'yes', 'no')
```

mfussenegger:
I wasn't sure what's the better tradeoff, having a miss-leading error message when looking at the `boolean` method in isolation (saying the range of choices is smaller than what it actually supports)

Having a inconsistency between the defined `choices` and the choices listed in the error message.

Or overloading the number of choices in the help output by setting `choices` to all supported inputs.

I'm even inclined to do a breaking change to limit it to (`true`, `false`, `yes`, `no`) and then have it consistent in both places.

PR #291, opened 2019-07-12
Update CrateDB used in tests to 4.0.2

mfussenegger:


chaudum:
I guess we need to update the Travis CI configuration so it uses JDK 12

mfussenegger:
First commit is also in a separate PR: https://github.com/crate/crash/pull/295

PR #293, opened 2019-07-15
Rename some methods to clarify their behavior

mfussenegger:
See individual commits


mfussenegger:
First commit is also in a separate PR: https://github.com/crate/crash/pull/295

PR #297, opened 2019-08-01
Change directory structure, switch from zope.testrunner to unittest

mfussenegger:
Main motivation is to be able to run tests from within an editor
directly using something like `vim-test`

Due to the `self.crate_host` hack that was in the `test_command` file,
those tests couldn't be run standalone as they required the monkey
patching done in the `test_suite` method.

It also gets rid of a few warnings caused by the older zope testing
version.


.travis.yml
```diff
@@ -19,11 +19,10 @@ before_install:
 
 install:
     - python bootstrap.py
-    - bin/buildout install scripts crate test linter
+    - bin/buildout install scripts crate linter
 
 script:
-    - bin/flake8
```
mfussenegger: Removed this for now because it's somehow messed up and checks files that shouldn't be checked:

```
bin/flake8
493/home/travis/build/crate/crash/bootstrap.py:146:5: E123 closing bracket does not match indentation of opening bracket's line
494/home/travis/build/crate/crash/docs/conf.py:1:1: F403 'from crate.theme.rtd.conf.crash import *' used; unable to detect undefined names
495/home/travis/build/crate/crash/devtools/bundle.py:24:31: E226 missing whitespace around arithmetic operator
```

The version is also outdated and doesn't catch some other issues.

Locally it also checks the `eggs` and `parts` folders, which yields tons of errors. Probably because of the buildout `sys.path` manipulation.

My plan is to later on get rid of buildout, and then I'll re-add it

tox.ini
```diff
@@ -9,4 +9,4 @@ deps =
     zope.testrunner
```
chaudum: So this line can be removed as well then.

PR #303, opened 2019-09-23
Drop Python 3.4 support

mfussenegger:
Some of the newer versions of our dependencies dropped support for
Python 3.4.

Python 3.4 has reached end of life and according to
https://en.wikipedia.org/wiki/CPython#Enterprise_Linux only Debian 8
shipped Python 3.4 to begin with.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup(
         ],
         argcompletion=['argcomplete']
     ),
-    python_requires='>=3.4',
+    python_requires='>=3.5',
     install_requires=requirements,
     package_data={'': ['*.txt']},
     classifiers=[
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -87,9 +87,9 @@ def read(path):
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
chaudum: Should we then also add 3.7 to the Travis CI configuration?
https://github.com/crate/crash/pull/303/files#diff-354f30a63fb0907d4ad57269548329e3R5

If so, I would also mention it in the changelog entry.

PR #304, opened 2019-10-01
Display cluster name in status line

chaudum:

You can see at a glance which cluster you're connected to.

## Checklist


chaudum:
> This will lead to a `[SchemaUnknownException: Schema 'sys' unknown` error if the user doesn't have sufficient permissions.

Right, did not think about that.

PR #308, opened 2019-10-21
Sort the output of the `\dt` command alphabetically

chaudum:
## Checklist


chaudum:
> +1
> I think this is worth a changes entry.

There is already an entry :)

seut:
@chaudum 
> There is already an entry :)

right, 🙄 my fault, I'll get some coffee 😉

PR #309, opened 2019-10-28
Update config module name

mechanomi:
the upstream name has changed in most recently docs theme package

- [ ] add min version for theme (https://github.com/crate/crash/pull/309#discussion_r339545930)
- [ ] sort backports (https://github.com/crate/crate/pull/9302#issuecomment-546934088)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,4 @@
-from crate.theme.rtd.conf.crash import *
+from crate.theme.rtd.conf.clients_crash import *
 
 site_url = 'https://crate.io/docs/clients/crash/en/latest/'
 extensions = ['sphinx_sitemap']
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -1,4 +1,4 @@
-from crate.theme.rtd.conf.crash import *
+from crate.theme.rtd.conf.clients_crash import *
```
chaudum: Does this require a certain minimum version of the docs theme? If so, please add the minimum version to the requirements.

mechanomi: good catch. I'll hold off until I do the release on the theme and then come back to this and take care of it

also, I need to take care of this: https://github.com/crate/crate/pull/9302#issuecomment-546934088

mechanomi:
waiting for crate/crate-docs-theme#162

docs/conf.py
mechanomi: okay fixed this now

mechanomi:
@chaudum can you re-review please? thanks!

PR #313, opened 2020-02-10
WIP: implement heartbeat to keep connection alive

chaudum:

@autophagy @mfussenegger Is the keep-alive feature of crash still a thing? It has been lying around on my local checkout for some time now and I am wondering whether I should get it over the finish line.

## Checklist

### DIFF
--- a/crate/crash/command.py
+++ b/crate/crash/command.py
@@ -28,12 +28,14 @@ import logging
 import os
 import re
 import sys
+import time
 from argparse import ArgumentParser, ArgumentTypeError
 from collections import namedtuple
 from distutils.version import StrictVersion
 from getpass import getpass
 from logging import NullHandler
 from operator import itemgetter
+from threading import Thread
 
 import urllib3
 from appdirs import user_config_dir, user_data_dir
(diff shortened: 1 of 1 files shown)


mfussenegger:
I'm not sure. `psql` suffers from the same problem when using it to connect to a cloud cluster. So I wonder if we can come up a solution that doesn't involve patching all clients.

chaudum:
> I'm not sure. `psql` suffers from the same problem when using it to connect to a cloud cluster. So I wonder if we can come up a solution that doesn't involve patching all clients.

The very specific problem, however, occurs when using Azure Loadbalancer, which has an idle timeout of 4-30 minutes, but it is not possible to deactivate connection reset.
https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-tcp-idle-timeout#tcp-idle-timeout
load-balancer/load-balancer-tcp-reset

mfussenegger:
Instead of doing query-heartbeats we could also enable TCP-keep-alive.

[urllib3 docs has an example how to set that](https://urllib3.readthedocs.io/en/latest/reference/#urllib3.connection.HTTPConnection)

chaudum:
> Instead of doing query-heartbeats we could also enable TCP-keep-alive.
> 
> [urllib3 docs has an example how to set that](https://urllib3.readthedocs.io/en/latest/reference/#urllib3.connection.HTTPConnection)

I will open an PR on crate-python.

PR #322, opened 2020-03-16
Dynamically load keywords from pg_get_keywords

autophagy:

Previously, our list of keywords we highlighted on was hard coded, and not tied to a version of CrateDB. This replaces that list, instead using the list of keywords returned by `pg_get_keywords()`

## Checklist

### DIFF
--- a/crate/crash/repl.py
+++ b/crate/crash/repl.py
@@ -111,83 +111,55 @@ class TruncatedFileHistory(FileHistory):
 
 
 class SQLCompleter(Completer):
-    keywords = [
-        "abs", "absolute", "action", "add", "after", "alias", "all", "allocate",
-        "alter", "always", "analyzer", "and", "any", "are", "array", "array_agg",
-        "array_max_cardinality", "as", "asc", "asensitive", "assertion",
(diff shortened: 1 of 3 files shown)


crate/crash/repl.py
```diff
@@ -111,83 +111,17 @@ def append(self, string):
...
     def __init__(self, cmd):
         self.cmd = cmd
+        self.keywords = self._populate_keywords()
+
+    def _populate_keywords(self):
+        success = self.cmd._exec("SELECT word FROM pg_catalog.pg_get_keywords()")
+        if success:
+            return [i[0] for i in self.cmd.cursor.fetchall()]
+        else:
+            return []
```
autophagy: @mfussenegger I'm not so sure about returning an empty list of the `pg_get_keywords()` call fails. Do you think a fallback list of a few most common SQL keywords would be better?

mfussenegger: > Do you think a fallback list of a few most common SQL keywords would be better?

Yes I think that would be good to have, for people who use a new crash to connect to old CrateDB clusters.

PR #340, opened 2020-09-28
Upgrade CrateDB python driver to 0.26.0

autophagy:
@chaudum As I understand, the TCP keepalive is enabled by default, so just upgrading the python driver should add it. Do you think it would be worth explicitly setting it to true for clarity's sake?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,12 @@ Changes for crash
 Unreleased
 ==========
 
+Changes
+=======
+
+- Upgraded the CrateDB python driver to 0.26.0 in order to enable TCP keepalive
+  on the socket level.
+
 2020/09/21 0.26.0
 =================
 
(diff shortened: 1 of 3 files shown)


chaudum:
> @chaudum As I understand, the TCP keepalive is enabled by default, so just upgrading the python driver should add it. Do you think it would be worth explicitly setting it to true for clarity's sake?

Right, keepalive is enabled by default, but depending on the OS level socket settings, this may not be enough, e.g. to mitigate the Azure networking issues we encountered.

I think it would make sense to set the socket options to values that[ make sense for use](https://social.msdn.microsoft.com/Forums/en-US/0462f436-2493-4199-ba1b-e6d05fdaebd4/azure-vm-tcp-idle-timeout?forum=WAVirtualMachinesforWindows#1cca8e0d-43a8-4c15-a946-12de725f2934), e.g.:

```
net.ipv4.tcp_keepalive_time = 120
net.ipv4.tcp_keepalive_intvl = 30
net.ipv4.tcp_keepalive_probes = 8
```

chaudum (approved):

PR #341, opened 2020-09-29
Upgrade to new docs build 0.4.0

mechanomi:
the primary purpose of this is to upgrade to the new docs build system

I noticed that the Travis config was gone but I can't find the commit that removed it. hmm

we currently use Travis for testing the docs. I've added back in a Travis config that only tests the docs

I've also gone ahead and added the Mergify config we're using on other docs projects. but I have slightly modified it to account for the GitHub Actions you have set up for testing the code

lmk what you think!! :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 .coverage
-.crate-docs-build
+.crate-docs
 .idea/
 .installed.cfg
 .mypy_cache/
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,25 @@
+name: docs
+on:
+  pull_request: ~
+  push:
+    branches:
+      - master
+
+jobs:
+  documentation:
+    name: Build docs on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Build docs
+        run: |
+          cd docs && make check
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,15 @@
+---
+pull_request_rules:
+  - actions:
+      merge:
+        method: rebase
+        rebase_fallback: null
+        strict: true
+    conditions:
+      - label=ready-to-merge
+      - '#approved-reviews-by>=1'
+      - status-success~=Test python
+      - status-success~=Build & publish package to pypi
+      - status-success~=Build docs
+      - status-success~=docs/readthedocs.org
+    name: default
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -26,8 +26,8 @@
 DOCS_DIR      := docs
 TOP_DIR       := ..
 BUILD_JSON    := build.json
-BUILD_REPO    := https://github.com/crate/crate-docs-build.git
-CLONE_DIR     := .crate-docs-build
+BUILD_REPO    := https://github.com/crate/crate-docs.git
+CLONE_DIR     := .crate-docs
 SRC_DIR       := $(CLONE_DIR)/src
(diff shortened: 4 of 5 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 .coverage
 .crate-docs-build
```
chaudum, line 2: I think you could remove that line, in favour of the new one.

chaudum:
> the primary purpose of this is to upgrade to the new docs build system
> 
> I noticed that the Travis config was gone but I can't find the commit that removed it. hmm

The commit was 243e83db11d3c39c6084b9409b521cd9f5e3d94a

> we currently use Travis for testing the docs. I've added back in a Travis config that only tests the docs

Sounds good to me!

> I've also gone ahead and added the Mergify config we're using on other docs projects. but I have slightly modified it to account for the GitHub Actions you have set up for testing the code

The introduction of mergify to this project isn't really related to the docs change, is it? Therefore I would make at least a separate commit for it.
 
> lmk what you think!! :)

mechanomi:
@chaudum to copy what I said via pm for visibility:

> I decided to include the mergify stuff because I wanted to lump all the changes to the build and CI/CD together because I'm already doing this across like 12 different projects so it's already quite exhausting
> is it okay in this instance to lump it together? I can alter the commit message if you like?
> re the .gitignore, I decided to leave it in so that people who used the old version don't accidentally commit the old build cache. my plan is to remove these entries at some point in the future. thoughts?

mfussenegger (changes requested):
Please don't add travis back.

We should move forward and create a github action for the docs build. 

Also, is it really necessary to build the docs both on osx and on linux?

mechanomi:
@mfussenegger as commented elsewhere: at the moment, we almost have all docs projects in lock-step wrt how they are built and how CI/CD is done. I would prefer that we only take steps forward, not steps backward (i.e., removing Travis brings this out of sync with the status quo)

there has been some talk of moving to GitHub Actions for testing the docs, but my current understanding is that we're having issues with limits. is there any concern with adding yet more GitHub Actions to our projects?

as soon as one GitHub Action configuration has been written, I will be more than happy to copy it across and roll it out to the rest of the docs projects. but I haven't found the time to learn how to do that yet. I asked Haudi on another PR if he would be able to do this but I haven't gotten a response to that yet (see https://github.com/crate/croud/pull/298)

as for why we're testing on Mac and Linux: because I want to be able to say that we support the docs build (for contributors) on both of these platforms, and we can't say that without actually testing it. there are some quite substantial differences in the way that stock Python is set up on both sytems

mfussenegger:
> there has been some talk of moving to GitHub Actions for testing the docs, but my current understanding is that we're having issues with limits. is there any concern with adding yet more GitHub Actions to our projects?

The limits only apply for private projects. Public repositories don't count towards it.

> as soon as one GitHub Action configuration has been written,

Something like the following should do the trick:

```
diff --git a/.github/workflows/docs.yml b/.github/workflows/docs.yml
new file mode 100644
index 0000000..9acc441
--- /dev/null
+++ b/.github/workflows/docs.yml
@@ -0,0 +1,24 @@
+---
+name: docs
+on:
+  pull_request: ~
+  push:
+    branches:
+      - master
+
+jobs:
+  documentation:
+    name: Build docs on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Build docs
+        run: |
+          cd docs && make qa && make check
```

mechanomi:
@mfussenegger omg thank you. I'll give it a shot on this PR when I get a bit of time to come back to this work

mechanomi:
@chaudum I have pushed fixups. can you please re-review? thx! <3

this PR now:

- includes a working GitHub Actions workflow for the docs
- adds a Mergify configuration
- pins the docs build system to 0.4.0

chaudum (approved):
lgtm

mfussenegger (approved):

PR #347, opened 2020-11-30
Fix "Sitemap" value in robots.txt

amotl:
This fixes the "Sitemap" value in `robots.txt`.

### DIFF
--- a/docs/_extra/robots.txt
+++ b/docs/_extra/robots.txt
@@ -1,4 +1,4 @@
 User-agent: *
 Disallow: /
 
-Sitemap: https://crate.io/docs/crash/crash/en/latest/site.xml
+Sitemap: https://crate.io/docs/crate/crash/en/latest/site.xml


mechanomi (approved):
good catch!

amotl:
Thanks for the review. What is this `Rule: default (merge) Expected — Waiting for status to be reported` the PR checks are waiting for?

mechanomi:
not sure. I don't have admin access to the branch protection config. perhaps @chaudum could help?

mechanomi:
wait. this is from https://github.com/crate/crash/blob/master/.mergify.yml

it looks like it was waiting for all the conditions to pass. the config looks correct, but maybe it was because the branch needed updating or something

mechanomi:
still. it says it's waiting for all the branch protection checks to pass. but they have passed afaik

try pushing a no-op to see if it fixes the situation?

i.e., `git rebase -i origin/master`, then select your commit for `r` (reword), but don't change the commit message, then force push. that should trigger the tests again

amotl:
https://github.com/crate/crash/pull/347/checks?check_run_id=1479973512 says:
> #### Branch protection settings are not validated anymore
>
> Branch protection is enabled and is preventing Mergify to merge the pull request. Mergify will merge when branch protection settings validate the pull request once again. (detail: Required status check "Rule: default (merge)" is cancelled.)

Do we really need to such an amount of protection here? I believe it might create more friction than not.

seut:
@Mergifyio refresh

mechanomi:
looks like a @crate/sysadmins needs to review the branch protections here. I cannot even see them to attempt to figure out what might be causing this :(

mfussenegger:
For some reason the mergify rule itself was added as a branch protection. No idea who did that. I removed it as it doesn't make sense to have that.

amotl:
Thanks!

PR #348, opened 2020-12-17
Correction: Use HTTPS instead of HTTP for links

mechanomi:

see crate/tech-writing-domain#316


mkleen (approved):
👍

Issue #349, opened 2020-12-18
Support executing multiple statements


amotl:
Hi there,

while I recognized #223 and #300 by @mfussenegger, I just found pasting of multiple statements would still (again?) not work on the `crash` REPL.

In general, this topic should not only be about pasting them into the UI/REPL, it is also about running multiple statements from STDIN. @jayeff provided a mcve quickly - thanks!
```
$ echo 'select 1; select 1;' | crash
CONNECT OK
SQLParseException[line 1:11: mismatched input 'select' expecting <EOF>]
```

With kind regards,
Andreas.

P.S.: I am running macOS with
```
$ crash --version
0.26.0
```

amotl:
It looks like splitting multiple SQL statements using newlines works.

### From file
```
$ cat test.sql
select 1;
select 1;
```

```
$ cat test.sql | crash
```

### Ad hoc
Make sure to use GNU's echo instead of BSD's echo.
```
$ gecho -e "select 1;\nselect 1;" | crash
```

amotl:
Starting with 0.24.2, the commit 369f2456 removed the line splitting in `crate.crash.command.process`.
```patch
-            for statement in _parse_statements(text.split('\n')):
+            for statement in _parse_statements([text]):
```

This has been coming from #274, but I am not able to find a specific use-case where the behavior before made things go haywire. @mxm said:
> Removing newlines has some unwanted side effects like line comments not terminating at the end of the line anymore.

Swapping those lines again on my end allows pasting of multiple lines into the UI again.
```
$ crash
CONNECT OK
cr> CREATE TABLE IF NOT EXISTS testdrive_ok ("name" STRING) CLUSTERED INTO 1 SHARDS;
    INSERT INTO testdrive_ok ("name") VALUES ('foobar');
    COPY testdrive_ok TO DIRECTORY '/tmp' WITH (format='json_object');

CREATE OK, 0 rows affected  (0.007 sec)
INSERT OK, 1 row affected  (0.051 sec)
COPY OK, 1 row affected  (0.041 sec)
```

Maybe @mxm or @mfussenegger can remember the reasons for this change again?

amotl:
Hi again,

my proposal GH-350 was only half-assed, and it does not make sense to implement it that way, so I am closing this patch. However, the discussion includes valuable information about the likely solution at https://github.com/crate/crash/pull/350#issuecomment-748878550 ff.

With kind regards,
Andreas.

PR #223, opened 2017-09-04
Support pasting of multiple statements

mfussenegger:
crash

PR #350, opened 2020-12-18
Fix execution when pasting multiple SQL statements into the REPL

amotl:
Hi there,

this topic has been tweaked within #223, #274 and #300 in the past. Now, we are revisiting it again within #349.

This commit only adjusts splitting multiple statements separated by the newline character "\n" again. It might be followed up by further adjustments for also splitting SQL statements on the boundaries of the semicolon character ";".

Let's check how CI will respond to that amendment.

With kind regards,
Andreas.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,7 +10,15 @@ Changes
 
 - Upgrade the CrateDB python driver to 0.26.0 in order to enable TCP keepalive
   on the socket level.
-- Make ctrl-d only quit program on empty command line, like any other shell
+- Make ctrl-d only quit program on empty command line, like any other shell.
+- Add optional ``--multiline`` command line argument.
+
+  When running with that option enabled, ``crash`` will accept pasting
+  multiline SQL statements on the REPL.
+
+  Multiline SQL statements in this context mean that multiple statements
+  separated by ``;<CR>`` will be accepted and will get invoked subsequently.
+  A single SQL statement is still allowed to span multiple lines as well.
 
 2020/09/21 0.26.0
 =================
--- a/crate/crash/command.py
+++ b/crate/crash/command.py
@@ -156,6 +156,10 @@ def get_parser(output_formats=[], conf=None):
                         default=_conf_or_default('format', 'tabular'),
                         choices=output_formats, metavar='FORMAT',
                         help='the output FORMAT of the SQL response')
+    parser.add_argument('--multiline', action='store_true',
+                        dest='multiline',
+                        default=_conf_or_default('multiline', False),
+                        help='enable multiline support')
     parser.add_argument('--version', action='store_true', default=False,
                         help='print the Crash version and exit')
 
@@ -211,6 +215,7 @@ class CrateShell:
                  is_tty=True,
                  autocomplete=True,
                  autocapitalize=True,
+                 multiline=False,
                  verify_ssl=True,
                  cert_file=None,
                  key_file=None,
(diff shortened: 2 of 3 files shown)


amotl:
> Let's check how CI will respond to that amendment.

While it does not fail on SQL statements spanning multiple lines per se, it croaks on SQL statements spanning multiple lines which define UDF functions. Bummer.

https://github.com/crate/crash/blob/8e4b6d0e6efdf76cc64ad291115c2c8c68ed7dc8/tests/test_integration.py#L667-L676

Most probably, @mfussenegger tried to tackle that edge case within 280a2e4, where he stated:
> ```
> Add udf function body handling to REPL multi-line detection logic
>
> If we're in a function body of a UDF definition, a `;<CR>` should not trigger the command.
> ```

However, the corresponding PR #300 was closed.

amotl:
I've added an option `--multiline` to enable accepting multiple SQL statements on the REPL. Feel free to add suggestions about a better naming for that option.

When `crash` will gain an SQL parser which can handle both multiple SQL statements and UDF functions at the same time, we might be able to remove that option again.

### Synopsis
```
$ crash --multiline
```

Copy & paste:
```
-- Easy
SELECT 1;

-- Math
SELECT 40+2;
```

chaudum:
Hi @amotl 
Thanks for looking again at multiline input support for crash.

Did you also consider using [sqlparse](https://pypi.org/project/sqlparse/) for splitting/parsing the input SQL statements? That would mitigate the problem of the semicolon character inside UDFs.

`$ cat multiline.py`:

```py

import sqlparse
import sys

def main():
    stmts = sqlparse.split(sys.stdin)
    for stmt in stmts:
        print("--------------")
        print(stmt)

main()
```

`$ cat udf.sql`:

```sql
SELECT 1;
CREATE OR REPLACE FUNCTION fib(LONG)
RETURNS LONG
LANGUAGE JAVASCRIPT
AS 'function fib(i) {
  if (i == 0) return 0;
  else if (i == 1) return 1;
  else return fib(n - 1) + fib(n - 2);
}';
-- comment
SELECT 1;
```

```console
$ python multiline.py < udf.sql
--------------
SELECT 1;
--------------
CREATE OR REPLACE FUNCTION fib(LONG)
RETURNS LONG
LANGUAGE JAVASCRIPT
AS 'function fib(i) {
  if (i == 0) return 0;
  else if (i == 1) return 1;
  else return fib(n - 1) + fib(n - 2);
}';
--------------
-- comment
SELECT 1;
```

amotl:
Hi @chaudum,

> Did you also consider using `sqlparse` for splitting/parsing the input SQL statements?

Not specifically, but yes: When touching this topic, I also thought that there should be a mature SQL parsing library for Python which already does the right thing in any condititions.

So, thanks a stack for sharing `sqlparse`. I believe we should replace the hand-rolled SQL parsing routines within one of the next iterations.

With kind regards,
Andreas.

mfussenegger:
I closed https://github.com/crate/crash/pull/300 because I realized that it doesn't work for all cases.

I don't really like adding a dedicated `--multiline` parameter, I think copy & pasting of multiple statements should "just work". I second @chaudum input that we should probably go with `sqlparse` for the line-splitting logic (or copy & vendor the relevant logic)

amotl:
Thanks.

> I don't really like adding a dedicated `--multiline` parameter.

I didn't like it as well but it was the only thing I have been able to quickly whip up.

> I second @chaudum input that we should probably go with sqlparse for the line-splitting logic.

Me too!

amotl:
Hi again,

thanks all for your valuable information and feedback on this matter. We will use it to implement a solution for the problem. In the meanwhile, let me refer dear readers back to GH-349, and close this one.

With kind regards,
Andreas.

Issue #343, opened 2020-10-19
Problem invoking the test suite


amotl:
Hi there,

within #342, I found that I haven't been able to invoke the test suite right away. In order to find out about more details, I tried again and would like to report the outcome here.

While I believe the process hasn't been able to download CrateDB 4.2.0 on the first try, now I am seeing:
```
$ python -m unittest -vvvvv

[...] some OKs

HTTP: http://127.0.0.1:44209
./var/folders/06/w9pzygdj7vx53n_0l9q_lhph0000gn/T/tmp2k7qhukz/crate-4.2.0/bin/crate: line 129: /var/folders/06/w9pzygdj7vx53n_0l9q_lhph0000gn/T/tmp2k7qhukz/crate-4.2.0/jdk/Contents/Home/bin/java: No such file or directory
/var/folders/06/w9pzygdj7vx53n_0l9q_lhph0000gn/T/tmp2k7qhukz/crate-4.2.0/bin/crate: line 129: exec: /var/folders/06/w9pzygdj7vx53n_0l9q_lhph0000gn/T/tmp2k7qhukz/crate-4.2.0/jdk/Contents/Home/bin/java: cannot execute: No such file or directory

[...] some more OKs

ERROR: setUpModule (tests.test_integration)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/amo/dev/crate/sources/crash/tests/test_integration.py", line 46, in setUpModule
    node.start()
  File "/Users/amo/dev/crate/sources/crash/.venv/lib/python3.8/site-packages/crate/testing/layer.py", line 316, in start
    self._wait_for_start()
  File "/Users/amo/dev/crate/sources/crash/.venv/lib/python3.8/site-packages/crate/testing/layer.py", line 373, in _wait_for_start
    self._wait_for(validator)
  File "/Users/amo/dev/crate/sources/crash/.venv/lib/python3.8/site-packages/crate/testing/layer.py", line 352, in _wait_for
    raise SystemError('Failed to start Crate instance in time.')
SystemError: Failed to start Crate instance in time.
```

It is probably just a minor issue. I am working on macOS and invoked that within a dedicated virtualenv after installing the package into it with
```
pip install -U -e ".[test]"
```

With kind regards,
Andreas.

mfussenegger:
This is because of the test layer https://github.com/crate/crash/blob/master/tests/test_integration.py#L39 loading the linux CrateDB tarball.

[...]

PR #351, opened 2020-12-19
Improve CI tests on GHA

amotl:
Hi there,

this patch slightly adjusts the CI/GHA configuration.

- Invoke integration tests on Linux only
  I found a quick way to disabled integration tests on all other operating systems than Linux. When we have solved #343, we will be able to run them on all operating systems.
- Add macOS, Windows and Python 3.9 to test matrix.

With kind regards,
Andreas.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,8 +8,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
-        python-version: [3.6, 3.7, 3.8]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, macos-latest, windows-2019]
```
amotl, line 11: I found that using `windows-2019` instead of `windows-2022`, which happened to become `windows-latest` just recently, significantly improves runtime, so I also submitted https://github.com/crate/crate/pull/12344.

.github/workflows/main.yml
```diff
@@ -21,13 +21,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install ".[test,devel]"
+          python -m pip install ".[test]"
 
       - name: Test
-        run: |
-          isort --recursive --check --diff crate/ tests/ setup.py
-          flake8 crate/crash
```
mfussenegger, line 29: Please keep these.

mfussenegger (approved):

PR #359, opened 2021-01-20
CrateDB bundles the JRE with recent releases so we don't need it on CI

amotl:
This removes installing Java on CI/GHA. Because CrateDB now bundles the JRE, this should not do any harm but will save some cycles instead.


seut (approved):
👍

PR #365, opened 2022-03-25
Update documentation subsystem, CI dependencies and Mergify configuration

amotl:
Hi there,

this is a maintenance patch which does three things.

- Update documentation builder subsystem to the most recent version
- Bump some versions of recipes within the CI configuration
- Update the Mergify configuration to the new schema

It was needed to verify configuration changes coming in from https://github.com/crate/crate-docs-theme/pull/328.

With kind regards,
Andreas.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,9 +17,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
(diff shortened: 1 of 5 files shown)


seut (approved):
👍 
A nitpick note: I'd prefer having a bit more context inside the commit messages. E.g. instead of just saying "Update mergify config" a note about the why would be cool, like 

Update mergify configuration
Use the new queue action instead of the deprecated merge action. 

That would help reviewers to understand what this is all about.

PR #367, opened 2022-04-12
Fix tabular output for rows containing only whitespace and friends

### LINKED ISSUE #366: text columns with whitespaces are not printed properly
Behaviour from Crash (0.27):
```
cr> create table t (a text);
CREATE OK, 1 row affected  (0.203 sec)
cr> insert into t values ('');
INSERT OK, 1 row affected  (0.039 sec)
cr> select * from t;
+---+
| a |
+---+
|   |
+---+
SELECT 1 row in set (0.006 sec)
cr> insert into t values (' ');
INSERT OK, 1 row affected  (0.015 sec)
cr> select * from t;
+---+
| a |
+---+
|   | // <-- 1 row printed
+---+
SELECT 2 rows in set (0.006 sec)
```

Expected (from psql connected to CrateDB):

```
crate=> create table t (a text);
CREATE 1
crate=> insert into t values ('');
INSERT 0 1
crate=> select * from t;
 a 
---
 
(1 row)

crate=> insert into t values (' ');
INSERT 0 1
crate=> select * from t;
 a 
---
    <-- 
    <-- 2 rows
(2 rows)
```

amotl:
Hi.

This patch intends to resolve #366. The issue is not present in the upstream [`tabulate` package](https://pypi.org/project/tabulate/), so it aims to integrate the upstream version and get rid of the vendored one. Instead, `tabulate.py` now contains a small monkeypatch to adjust some bits we need there [A].

A single test case had to be adjusted where column values changed their alignment [B]. Do you think it will be okay?

With kind regards,
Andreas.

----

### A. Gist of the monkeypatch

https://github.com/crate/crash/blob/6fca4c6103baa7d1f5d6b0dea3725421fab659ad/crate/crash/tabulate.py#L51-L54

Originally, this was coming from 1052e0d79. Apparently, it is the only patch needed to make the upstream version compatible with the previously vendored one. It looks pretty innocent [A.1], maybe there is a chance to bring it into mainline `tabulate` on another iteration.

[A.1] But it is definitively needed. Otherwise, the very test case outlined below would croak, IIRC.

### B. Adjustment to test case `test_pprint_dont_guess_type`

```diff
diff --git a/tests/test_integration.py b/tests/test_integration.py
index 97bb873..56f7d7c 100644
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -202,7 +202,7 @@ def test_pprint_dont_guess_type(self):
         expected = "\n".join(["+---------+",
                               "| version |",
                               "+---------+",
-                              "| 0.50    |",
+                              "|    0.50 |",
                               "+---------+\n"])
         with CrateShell() as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
```


crate/crash/tabulate.py
```diff
@@ -612,546 +29,46 @@ def _format(val, valtype, floatfmt, missingval="", has_invisible=True):
...
+        )
         if is_a_colored_number:
             raw_val = _strip_invisible(val)
             formatted_val = format(float(raw_val), floatfmt)
             return val.replace(raw_val, formatted_val)
+        # PATCH: Preserve string formatting even for numeric looking values.
+        # https://github.com/crate/crash/commit/1052e0d79
         elif not floatfmt:
-            return float_format(val)
+            return str(val)
```
amotl, line 54: Apparently, this is the only patch we need to make the upstream version compatible with the previously vendored one. It looks pretty innocent, maybe there is a chance to bring it into mainline?

mfussenegger (approved):

PR #369, opened 2022-04-26
Dropped support for Python < 3.7

hramezani:

As Python 3.5 and 3.6 don't receive updates anymore, I think we can drop their support.
## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-2019]
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10']
         cratedb-version: ['3.3.6', '4.7.1']
     env:
       OS: ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


amotl:
Dear Hasan,

thanks for your patch.

> As Python 3.5 and 3.6 don't receive updates anymore,

So true!

> I think we can drop their support.

I am not sure about this detail. I tend to be more conservative and keep support for older variants, even if EOL, when there is no other pressing reason to remove it.

For example, on another machine we are running, Debian 9 "stretch" still has Python 3.5.
```
root@quepasa:~# cat /etc/debian_version
9.13

root@quepasa:~# python3 -V
Python 3.5.3
```

With kind regards,
Andreas.

P.S.: Also pinging @seut to gather his opinion about it.

hramezani:
Dear Andreas,

Thanks for your reply.

I can see a lot of popular python packages like [urllib3](https://github.com/urllib3/urllib3), [Django](https://github.com/django/django), [Pytest](https://github.com/pytest-dev/pytest) already dropped older python versions support.
IMHO, we can push people to use modern python versions by dropping older python version support from our packages.

> I tend to be more conservative and keep support for older variants, even if EOL, when there is no other pressing reason to remove it.

True. 

I will leave this PR to you to decide about it. Feel free to close the PR if you want to keep the older python version support.

Best regards,
Hasan

hramezani:
Ok, Thanks for your response @seut. I close the PR.

PR #374, opened 2022-11-22
Chore: CI maintenance

amotl:
- Update GHA recipe versions.
- Update CrateDB versions, use 4.8.4 and 5.1.1.
- Add support for Python 3.11.


seut (approved):
👍 Could you follow up and upgrade also CrateDB used inside the test https://github.com/crate/crash/blob/master/tests/test_integration.py#L30? Thx!

PR #375, opened 2023-01-16
Chore: Regular maintenance updates

amotl:
Hi there,

this patch does not change too much, but accomodates the update to `isort5` with 7fb303c, and protects against major version bumps of dependency packages with 85ac5d989f.

While trying to make CI succeed, the CodeQL configuration had to be adjusted (844942b, 54d5882), and, while being at it, 7dff1e8 has been added to save resources when quickly pushing to CI consecutively.

With kind regards,
Andreas.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -51,7 +51,7 @@ jobs:
 
       - name: Install project
         run: |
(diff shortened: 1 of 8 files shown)


seut (approved):
👍

matriv (approved):

PR #376, opened 2023-01-17
Dependencies: Update to tabulate 0.9.0, and drop support for Python 3.5 and 3.6

amotl:
In the spirit of modernizing the package dependencies, in order to allow the program to co-exist within Python environments needing a more recent version of the `tabulate` package, this patch uses the gist of https://github.com/astanin/python-tabulate/pull/161 to accomodate the update.

~~P.S.: While that patch to `tabulate` officially drops support for Python 3.6, the `crash` test suite still succeeds, so why not keep it for a while.~~

Following that patch to `tabulate`, this patch implicitly drops support for Python 3.6 here. Nevertheless, we will state it within the changelog explicitly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,10 +59,6 @@ To update, run::
 
     pip install -U crash
 
-If you are not using Python version 3.5 or above, recent version of `pip`_ will
-install an earlier version of Crash. This is because newer versions of this
-package are not compatible with Python 2.7 or 3.4 and below.
-
 Standalone
 ----------
 
--- a/crate/crash/tabulate.py
+++ b/crate/crash/tabulate.py
@@ -1,4 +1,3 @@
-import tabulate
 from tabulate import DataRow, Line as TabulateLine, TableFormat, _strip_ansi
 
 crate_fmt = TableFormat(lineabove=TabulateLine("+", "-", "+", "+"),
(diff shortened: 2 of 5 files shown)


crate/crash/tabulate.py
```diff
@@ -59,6 +51,7 @@
 
 
 def monkeypatch():
+    import tabulate
```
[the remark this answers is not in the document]
amotl: I've dismissed this admonition, because the "double import" has its roots in this monkeypatch.

README.rst
```diff
@@ -59,16 +59,11 @@ To update, run::
...
-install an earlier version of Crash. This is because newer versions of this
-package are not compatible with Python 2.7 or 3.4 and below.
-
 Standalone
 ----------
 
 Crash is also available as a standalone executable that includes all the
-necessary dependencies, and can be run as long as Python (>= 3.5) is available
-on your system.
+necessary dependencies.
```
mfussenegger, line 66: Shouldn't we keep mentioning the minimal required python version for the bundles?

mfussenegger (approved):

README.rst
amotl: Oh, maybe I was too quick to remove this paragraph. I've just brought it back. Thanks for spotting.

Issue #338, opened 2020-09-09
Update prompt-toolkit dependency to 3.0.x


mfussenegger:
[prompt_toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit) 3.0 has been out for a while.

I think we could update the dependency

avats-dev:
I think I've to update this line https://github.com/crate/crash/blob/9177c001ce73eb80c295ccb9bc21ef23edbf72a7/setup.py#L33 , right @mfussenegger ? Should I open a PR with the changes?

mfussenegger:
Yes you've to update that line. Maybe it's also necessary to make some other adjustements depending on the changes within prompt-toolkit. 


PR would be nice :+1:

PR #379, opened 2023-01-17
Dependencies: Update to prompt-toolkit v3

amotl:
In the same spirit as the other recent patches, this one updates the `prompt-toolkit` dependency package. It will resolve GH-338.

### DIFF
--- a/crate/crash/repl.py
+++ b/crate/crash/repl.py
@@ -41,7 +41,7 @@ from prompt_toolkit.layout.processors import (
     ConditionalProcessor,
     HighlightMatchingBracketProcessor,
 )
-from prompt_toolkit.output.defaults import get_default_output
+from prompt_toolkit.output import create_output
 from prompt_toolkit.styles.pygments import style_from_pygments_cls
 from pygments.lexers.sql import PostgresLexer
 from pygments.style import Style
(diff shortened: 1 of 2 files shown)


crate/crash/repl.py
```diff
@@ -308,7 +308,7 @@ def loop(cmd, history_file):
         get_bottom_toolbar_tokens=lambda: get_toolbar_tokens(cmd),
         get_prompt_tokens=lambda: [('class:prompt', 'cr> ')]
     )
-    output = get_default_output()
+    output = create_output()
```
amotl, line 311: Do you think it would be sensible to use `create_output(always_prefer_tty=True)`? I tried it, and the test suite will still succeed.

```python
def create_output(
    stdout: Optional[TextIO] = None, always_prefer_tty: bool = False
) -> Output:
    """
    Return an :class:`~prompt_toolkit.output.Output` instance for the command
    line.

    :param stdout: The stdout object
    :param always_prefer_tty: When set, look for `sys.stderr` if `sys.stdout`
        is not a TTY. Useful if `sys.stdout` is redirected to a file, but we
        still want user input and output on the terminal.
        By default, this is `False`. If `sys.stdout` is not a terminal (maybe
        it's redirected to a file), then a `PlainTextOutput` will be returned.
        That way, tools like `print_formatted_text` will write plain text into
        that file.
```

-- https://github.com/prompt-toolkit/python-prompt-toolkit/blob/3.0.36/src/prompt_toolkit/output/defaults.py#L19-L35

mfussenegger: I don't think so

mfussenegger (approved):

seut (approved):

PR #383, opened 2023-02-16
CI: Reduce resource usage and speed up testing cycle

amotl:
I found the CI to be rather heavy, it takes ages to complete. This patch makes it a bit more lightweight, following practises we already applied to other repositories successfully [1], without ending up missing any incompatibility issues so far.

- ~~Take CrateDB 4.8 release out of the test matrix.~~
- Run only a single test matrix slot for macOS and Windows, using the most recent stable Python 3.11. I think it is acceptable.

#### References
- [1] https://github.com/earthobservations/wetterdienst/pull/817

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,9 +12,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        os: ['ubuntu-latest']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         cratedb-version: ['4.8.4', '5.2.2']
+        include:
+          - os: 'macos-latest'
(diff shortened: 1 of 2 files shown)


matriv (commented):
Not very sure about this, please let's wait for some other opinions here.

proddata:
I think this should at least also include a test with the last major CrateDB version, as there are still many clusters running 4.x versions. However it might be sufficient to do this on a single OS/python combo.

amotl:
I've updated the patch to follow the suggestion by @proddata, and re-enabled integration testing against CrateDB 4.8.4. Integration testing is currently only active/possible on Linux.

https://github.com/crate/crash/blob/1f2a82d230592787a501a08e06d6a1b60c3d520f/tests/test_integration.py#L28-L29

matriv (approved):
thx!

proddata (approved):

PR #386, opened 2023-06-11
Add support for multi-line SQL and commands.

### LINKED ISSUE #349: Support executing multiple statements

JustBeYou:
Fixes #349 

As discussed in the issue, I replaced the custom SQL parsing with `sqlparse` library as it seems to be more reliable. Also, I saw there is an inconsistency between `process` and `process_iterable` as one was able to execute commands and the other did not, so I refactored how they work a little.

Now, multi-line input can contain single-line SQL, multi-line SQL and commands, all together.

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -229,16 +229,24 @@ SELECT 1;
 -- Another SELECT statement.
 SELECT 2;
 -- Yet another SELECT statement with an inline comment.
--- Other than the regular comments, it gets passed through to the database server.
+-- Comments get passed through to the database server.
 SELECT /* this is a comment */ 3;
+SELECT /* this is a multi-line
+comment */ 4;
 """
         cmd = CrateShell()
         cmd._exec_and_print = MagicMock()
         cmd.process_iterable(sql.splitlines())
-        cmd._exec_and_print.assert_has_calls([
-            call("SELECT 1"),
-            call("SELECT 2"),
-            call("SELECT /* this is a comment */ 3"),
+        self.assertListEqual(cmd._exec_and_print.mock_calls, [
+            call("-- Just a dummy SELECT statement.\nSELECT 1;"),
+            call("-- Another SELECT statement.\nSELECT 2;"),
+            call('\n'.join([
+                    "-- Yet another SELECT statement with an inline comment.",
+                    "-- Comments get passed through to the database server.",
+                    "SELECT /* this is a comment */ 3;"
+                ])
+            ),
+            call('SELECT /* this is a multi-line\ncomment */ 4;')
         ])
 
     def test_js_comments(self):
@@ -262,3 +270,112 @@ SELECT /* this is a comment */ 3;
         cmd.process(sql)
         self.assertEqual(1, cmd._exec_and_print.call_count)
         self.assertIn("CREATE FUNCTION", cmd._exec_and_print.mock_calls[0].args[0])
+
+
+class MultipleStatementsTest(TestCase):
+
+    def test_single_line_multiple_sql_statements(self):
+        cmd = CrateShell()
+        cmd._exec_and_print = MagicMock()
+        cmd.process("SELECT 1; SELECT 2; SELECT 3;")
+        self.assertListEqual(cmd._exec_and_print.mock_calls, [
+            call("SELECT 1;"),
+            call("SELECT 2;"),
+            call("SELECT 3;"),
+        ])
+
+    def test_multiple_lines_multiple_sql_statements(self):
+        cmd = CrateShell()
+        cmd._exec_and_print = MagicMock()
+        cmd.process("SELECT 1;\nSELECT 2; SELECT 3;\nSELECT\n4;")
+        self.assertListEqual(cmd._exec_and_print.mock_calls, [
+            call("SELECT 1;"),
+            call("SELECT 2;"),
+            call("SELECT 3;"),
+            call("SELECT\n4;"),
+        ])
+
+    def test_single_sql_statement_multiple_lines(self):
+        """When processing single SQL statements, new lines are preserved."""
+
+        cmd = CrateShell()
+        cmd._exec_and_print = MagicMock()
+        cmd.process("\nSELECT\n1\nWHERE\n2\n=\n3\n;\n")
+        self.assertListEqual(cmd._exec_and_print.mock_calls, [
+            call("SELECT\n1\nWHERE\n2\n=\n3\n;"),
+        ])
+
+    def test_multiple_commands_no_sql(self):
+        cmd = CrateShell()
+        cmd._try_exec_cmd = MagicMock()
+        cmd._exec_and_print = MagicMock()
+        cmd.process("\\?\n\\connect 127.0.0.1")
+        cmd._try_exec_cmd.assert_has_calls([
+            call("?"),
+            call("connect 127.0.0.1")
+        ])
+        cmd._exec_and_print.assert_not_called()
+
+    def test_commands_and_multiple_sql_statements_interleaved(self):
+        """Combine all test cases above to be sure everything integrates well."""
+
+        cmd = CrateShell()
+        mock_manager = MagicMock()
+
+        cmd._try_exec_cmd = mock_manager.cmd
+        cmd._exec_and_print = mock_manager.sql
+
+        cmd.process("""
+    \\?
+    SELECT 1
+        WHERE 2 = 3; SELECT 4;
+    \\connect 127.0.0.1
+    SELECT 5
+        WHERE 6 = 7;
+    \\check
+        """)
+
+        self.assertListEqual(mock_manager.mock_calls, [
+            call.cmd("?"),
+            call.sql('SELECT 1\nWHERE 2 = 3;'),
+            call.sql('SELECT 4;'),
+            call.cmd("connect 127.0.0.1"),
+            call.sql('SELECT 5\nWHERE 6 = 7;'),
+            call.cmd("check"),
+        ])
+
+    def test_comments_along_multiple_statements(self):
+        """Test multiple types of comments along multi-statement input."""
+
+        cmd = CrateShell()
+        cmd._exec_and_print = MagicMock()
+
+        cmd.process("""
+-- Multiple statements and comments on same line
+
+SELECT /* inner comment */ 1; /* this is a single-line comment */ SELECT /* inner comment */ 2;
+
+-- Multiple statements on multiple lines with multi-line comments between them
+
+SELECT /* inner comment */ 3; /* this is a
+multi-line comment */ SELECT /* inner comment */ 4;
+
+-- Multiple statements on multiple lines with multi-line comments between and inside them
+
+SELECT /* inner multi-line
+comment */ 5 /* this is a multi-line
+comment before statement end */; /* this is another multi-line
+comment */ SELECT /* inner multi-line
+comment */ 6;
+        """)
+
+        self.assertListEqual(cmd._exec_and_print.mock_calls, [
+            call('-- Multiple statements and comments on same line\n\nSELECT /* inner comment */ 1;'),
+            call('/* this is a single-line comment */ SELECT /* inner comment */ 2;'),
+
+            call('-- Multiple statements on multiple lines with multi-line comments between them\n\nSELECT /* inner comment */ 3;'),
+            call('/* this is a\nmulti-line comment */ SELECT /* inner comment */ 4;'),
+
+            call('-- Multiple statements on multiple lines with multi-line comments between and inside them\n\nSELECT /* inner multi-line\ncomment */ 5 /* this is a multi-line\ncomment before statement end */;'),
+            call('/* this is another multi-line\ncomment */ SELECT /* inner multi-line\ncomment */ 6;')
+        ])
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,8 @@ Unreleased
 ==========
 
 - Fix inconsistent spacing around printed runtime. Thank you, @hammerhead.
+- Add support for multi-line input of commands and SQL statements for both
+  copy-pasting inside the crash shell and input pipes into crash.
 
 2023/02/16 0.29.0
 =================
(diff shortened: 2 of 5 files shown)


amotl:
Dear Mihail,

thanks a stack for your contribution, and apologies for the delay reviewing it.

I've just added a few maintenance patches, so you may have to rebase your patch after they have been integrated. Specifically, the improvements most-related to your patch are:

- GH-387 fixes a CI flaw, that's why software tests have not been invoked on your pull request.
- While reviewing your patch, @seut discovered that the test suite currently does not offer any tests including SQL comments. Because this may well become relevant for your improvements, I added GH-388 to improve coverage on that angle.

With kind regards,
Andreas.

tests/test_commands.py
```diff
@@ -217,3 +217,36 @@ def test_check_command_with_node_check(self, cmd):
...
+    def test_input_commands_amd_sql(self):
+        cmd = CrateShell()
+        cmd._try_exec_cmd = MagicMock()
+        cmd._exec_and_print = MagicMock()
+        cmd.process_iterable(["SELECT 1;\n\\?", "SELECT 2;"])
+        cmd._try_exec_cmd.assert_called_once_with("?")
+        cmd._exec_and_print.assert_has_calls([
+            call("SELECT 1;"),
+            call("SELECT 2;"),
+        ])
```
amotl, line 252: At https://github.com/crate/crash/pull/350#issuecomment-748878550, @chaudum outlined an SQL statement including a JavaScript function. With GH-388, I added software tests including SQL comments.

It will be sweet if a corresponding test case could be added here, to verify all together will start to work well after bringing in a real SQL parser.

- Multiline SQL statements including newline characters.
- Multiline JavaScript functions including newline characters.
- Multiple SQL statements.
- Comments in SQL and JavaScript.

JustBeYou: After #388 gets merged, I will rebase and enhance the testing suite to encompass all cases.

JustBeYou:
Thanks for your review, Andreas! I did not notice that the pipeline was not running, huh.

By the way, there is no mention of the linting tools in the development guide. I think it would be nice to to add a little script that runs all the important CI steps locally (linting and testing). Maybe a Git pre-commit hook for linting would be nice too.  If this sounds good to you, I'll open a new PR for that.

tests/test_commands.py
```diff
@@ -262,3 +262,37 @@ def test_js_comments(self):
...
```
amotl, line 298: I've verified the command-line examples at GH-349 start working with your patch, both when using newlines for separating SQL statements, or not. Excellent.

```console
echo -e "select 1; select 2;" | crash
echo -e "select 1;\nselect 2;" | crash
```

Can you make sure you are also exercising both variants within your test cases?

tests/test_commands.py
```diff
@@ -236,9 +236,9 @@ def test_sql_comments(self):
         cmd._exec_and_print = MagicMock()
         cmd.process_iterable(sql.splitlines())
         cmd._exec_and_print.assert_has_calls([
-            call("SELECT 1"),
-            call("SELECT 2"),
-            call("SELECT /* this is a comment */ 3"),
+            call("SELECT 1;"),
+            call("SELECT 2;"),
+            call("SELECT  3;"),
```
amotl, line 241: I actually expected that the `sqlparse` library would filter out any SQL comments, what this adjustment reflects well.

At [^1], I mentioned:

> I think it may be essential that inline SQL comments get forwarded 1:1, because some SQL dialects accept annotations/decorations through specific inline SQL comment syntaxes, right?

So, it may be advisable to use `sqlparse`'s `strip_comments=False` option [^2] instead, if that would be possible?

P.S.: I think CrateDB does not use inline SQL comments for any purpose. I am merely looking at this detail from a general perspective [^3].

[^1]: https://github.com/crate/crash/pull/388#issuecomment-1595813170
[^2]: https://sqlparse.readthedocs.io/en/latest/api/#formatting-of-sql-statements
[^3]: Please advise or share your thoughts about this aspect, @seut, @matriv, and @hammerhead.

amotl: > I think CrateDB does not use inline SQL comments for any purpose. I am merely looking at this detail from a general perspective.

Let me reference a fragment of a typical MySQL/MariaDB database dump file, created using the `mysqldump` command, which outlines corresponding inline SQL comments which are used to convey special commands/annotations/decorations.

<details>
<summary>mysqldump example</summary>

```sql
-- MySQL dump 10.16  Distrib 10.1.48-MariaDB, for debian-linux-gnu (i686)
--
-- Host: localhost    Database: test
-- ------------------------------------------------------
-- Server version	10.1.48-MariaDB-0+deb9u2

/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
/*!40101 SET NAMES utf8mb4 */;
/*!40103 SET @OLD_TIME_ZONE=@@TIME_ZONE */;
/*!40103 SET TIME_ZONE='+00:00' */;
/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
/*!40014 SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 */;
/*!40101 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='NO_AUTO_VALUE_ON_ZERO' */;
/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;

--
-- Table structure for table `ccms_messages`
--

DROP TABLE IF EXISTS `ccms_messages`;
/*!40101 SET @saved_cs_client     = @@character_set_client */;
/*!40101 SET character_set_client = utf8 */;
CREATE TABLE `ccms_messages` (
  `id` int(11) NOT NULL,
  `sender` varchar(32) NOT NULL,
  `recievers` varchar(255) DEFAULT NULL,
  `msg_subject` varchar(255) DEFAULT NULL,
  `msg_body` text,
  `timestamp` int(11) DEFAULT NULL,
  `is_composed` tinyint(1) DEFAULT NULL,
  `is_new` tinyint(1) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=MyISAM DEFAULT CHARSET=latin1;
/*!40101 SET character_set_client = @saved_cs_client */;

--
-- Dumping data for table `ccms_messages`
--

LOCK TABLES `ccms_messages` WRITE;
/*!40000 ALTER TABLE `ccms_messages` DISABLE KEYS */;
/*!40000 ALTER TABLE `ccms_messages` ENABLE KEYS */;
UNLOCK TABLES;
```

</details>

JustBeYou: Looking at [CrateDB's SQL Compatibility documentation](https://crate.io/docs/crate/reference/en/master/appendices/compatibility.html), I see you aim for PostgreSQL compatibility where such special comments do not exist as far as I know. 

If CrateDB does not use inline comments for any special purpose nor it tries to achieve MySQL/MariaDB full compatibility, I do not see a reason to keep sending the comments to the database.

I'm waiting for more input on this problem and I'll update the PR accordingly after.

tests/test_commands.py
JustBeYou: I have just added test cases to encompass the example you gave. I also improved the naming of existing test cases and added a few new ones. As a side note, I'm always testing with 3 statements instead of 2, because in my experience, a lot of things work with <= 2 elements, but not with more. :pain_emoji_needed:

tests/test_commands.py
```diff
@@ -262,3 +262,74 @@ def test_js_comments(self):
...
+        cmd.process("""
+    \\?
+    SELECT 1
+        WHERE 2 = 3; SELECT 4;
+    \\connect 127.0.0.1
+    SELECT 5
+        WHERE 6 = 7;
+    \\check
+        """)
+        cmd._try_exec_cmd.assert_has_calls([
```
JustBeYou, line 326: Maybe ordering should be somehow integrated into this testcase (eg. `\\?` gets called before `SELECT  1` statement), but I am not sure if it's worth it as it would require a [piece of code](https://stackoverflow.com/a/22677452) I do not consider elegant.

tests/test_commands.py
amotl: I agree with your assessment. Apologies for my suggestion, it was not appropriate.

> [Any kind of] comment is removed from the input stream before further syntax analysis and is effectively replaced by whitespace.
>
> -- https://www.postgresql.org/docs/8.0/sql-syntax.html#SQL-SYNTAX-COMMENTS

tests/test_commands.py
```diff
@@ -262,3 +262,74 @@ def test_js_comments(self):
...
+            call("?"),
+            call("connect 127.0.0.1"),
+            call("check")
+        ])
+        cmd._exec_and_print.assert_has_calls([
+            call('SELECT 1\nWHERE 2 = 3;'),
+            call('SELECT 4;'),
+            call('SELECT 5\nWHERE 6 = 7;')
+        ])
```
amotl, line 335: > Verifying the order of the calls would require a [piece of code](https://stackoverflow.com/a/22677452) I do not consider quite elegant.

They must be aiming at something more advanced with `attach_mock`, whatever this does. Actually, it only needs a slight adjustment, directly using the `mock_calls` attribute without further ado.

```suggestion
        self.assertListEqual(cmd._try_exec_cmd.mock_calls, [
            call("?"),
            call("connect 127.0.0.1"),
            call("check")
        ])
        self.assertListEqual(cmd._exec_and_print.mock_calls, [
            call('SELECT 1\nWHERE 2 = 3;'),
            call('SELECT 4;'),
            call('SELECT 5\nWHERE 6 = 7;')
        ])
```

JustBeYou: I was thinking more about verifying the ordering of SQL statements alongside commands, something like:
```
call("?"),
call('SELECT 1\nWHERE 2 = 3;'),
call("connect 127.0.0.1"),
call('SELECT 4;'),
call("check"),
call('SELECT 5\nWHERE 6 = 7;')
```

amotl: Ah, I see. I missed that there are two different mocks involved. Yeah, why not use such a Mock manager for that purpose?

amotl, line 335:
```suggestion
        manager = Mock()
        manager.attach_mock(cmd._try_exec_cmd, "cmd")
        manager.attach_mock(cmd._exec_and_print, "sql")
        cmd.process("""
    \\?
    SELECT 1
        WHERE 2 = 3; SELECT 4;
    \\connect 127.0.0.1
    SELECT 5
        WHERE 6 = 7;
    \\check
        """)
        self.assertListEqual(manager.mock_calls, [
            call.cmd("?"),
            call.sql("SELECT 1\nWHERE 2 = 3;"),
            call.sql('SELECT 4;'),
            call.cmd("connect 127.0.0.1"),
            call.sql('SELECT 5\nWHERE 6 = 7;'),
            call.cmd("check")
        ])
```

JustBeYou: I think it was looking uglier in my head, but I did not actually write it. Now, it seems alright. I've also changed the other test cases to account for ordering.

amotl (approved):
Thanks Mihail. The patch looks good to me. Please wait for the others' review before squashing the commits.

tests/test_commands.py
seut: Can you please adjust the comment [here](https://github.com/crate/crash/pull/386/files#diff-5592e672cef095f585fc0083c44cdfb8aafdfc11904a484dfa2b0794b0079f35R232) as inline comments are stripped out now?

amotl: Thanks for spotting!

amotl: Aha. At https://github.com/crate/crate/issues/13283, there is a proposal to indeed use SQL comments for sending hints to the execution engine. Maybe it is safer to use the `strip_comments=False` parser option right from the beginning?

JustBeYou: Well, I did not see that proposal. Taking it into account, `strip_comments=False` seems the right decision in order to avoid future inconveniences.

amotl: Hi Mihail,

@matriv just confirmed your assessment at https://github.com/crate/crate/issues/13283#issuecomment-1600468021. So, may I ask you to adjust the patch to use `strip_comments=False`? Thank you so much!

With kind regards,
Andreas.

matriv: So, can we please have some test(s) with inline, and multiline comments?
It's server-side job to treat them and shouldn't be stripped by crash.

amotl: Hi Marios. Thank you for chiming in. I added corresponding test cases on behalf of GH-388 already, to set the stage, and asked @JustBeYou to adjust his patch accordingly. So, I think we agree on those matters, and it will have a good outcome.

matriv: I know, just thought it would be nice to have also test with comments in the multi-line, multi-command context.

JustBeYou: As the conversation seems to have settled, I updated `strip_comments` to be false and added a few more test cases.
Notice that, `strip_comments=False` will keep all kinds of comments.

amotl (approved):
👍

seut (approved):
👍 thank you!

matriv (approved):
Thank you @JustBeYou !

amotl:
Dear Mihail,

your patch is well received. If you could add a changelog item, and squash your commits, I will integrate it, and run a new release. Thank you again.

With kind regards,
Andreas.

JustBeYou:
@amotl everything should be ready after the pipeline finishes. :rocket: Thank you

PR #388, opened 2023-06-17
Tests: Add test cases using comments within SQL and JS statements

amotl:
@seut discovered we did not have a single test case using SQL comments. This patch adds two corresponding test cases.

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -21,9 +21,10 @@
 
 import os
 import shutil
+import sys
 import tempfile
-from unittest import TestCase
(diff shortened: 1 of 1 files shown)


matriv (approved):
Thx, can we also have a comment within an sql line? like:
```
SELECT /* this is a comment */ 1;
```
and:
```
SELECT 1 -- this is a comment
;
```
(If we support this on cratedb side of course).

amotl:
> Can we also have a comment **within** an SQL statement?

Good idea. 01c153dc483437 adds a corresponding SQL statement.

Issue #384, opened 2023-03-29
Inconsistent spacing around printed runtime


hammerhead:
For some types of statements, there are two spaces between the message (`1 row affected`) and the runtime (`0.172 sec`) instead of one.

```
cr> SELECT 1;
+---+
| 1 |
+---+
| 1 |
+---+
SELECT 1 row in set (0.002 sec)
                   ^ one space
cr> CREATE TABLE a (b INTEGER);
CREATE OK, 1 row affected  (0.172 sec)
                          ^ two spaces
```

amotl:
Thank you for the report. The flaw has been fixed with GH-389, and the fix will be part of the next release.

PR #389, opened 2023-06-17
Fix inconsistent spacing around printed runtime

amotl:
This patch aims to fix GH-384 reported by @hammerhead.

> For some types of statements, there have been two spaces between the message (1 row affected) and the runtime (0.172 sec) instead of one.


matriv (approved):
thx!

PR #405, opened 2023-11-06
CI: Use CrateDB 5.4.5

amotl:
## About

On CI, use a more recent version of CrateDB.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,14 +26,14 @@ jobs:
       matrix:
         os: ['ubuntu-latest']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
-        cratedb-version: ['4.8.4', '5.3.2']
+        cratedb-version: ['4.8.4', '5.4.5']
         include:
           - os: 'macos-latest'
             python-version: '3.12'
-            cratedb-version: '5.3.2'
+            cratedb-version: '5.4.5'
           - os: 'windows-latest'
             python-version: '3.12'
-            cratedb-version: '5.3.2'
+            cratedb-version: '5.4.5'
     env:
       OS: ${{ matrix.os }}
       PYTHON: ${{ matrix.python-version }}


.github/workflows/main.yml
```diff
@@ -26,14 +26,14 @@ jobs:
       matrix:
         os: ['ubuntu-latest']
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
-        cratedb-version: ['4.8.4', '5.3.2']
+        cratedb-version: ['4.8.4', '5.4.4']
```
seut, line 29: Same here, lets use `5.4.5`?
```suggestion
        cratedb-version: ['4.8.4', '5.4.5']
```

.github/workflows/main.yml
```diff
@@ -26,14 +26,14 @@ jobs:
...
         include:
           - os: 'macos-latest'
             python-version: '3.12'
-            cratedb-version: '5.3.2'
+            cratedb-version: '5.4.4'
```
seut, line 33:
```suggestion
            cratedb-version: '5.4.5'
```

.github/workflows/main.yml
```diff
@@ -26,14 +26,14 @@ jobs:
...
           - os: 'windows-latest'
             python-version: '3.12'
-            cratedb-version: '5.3.2'
+            cratedb-version: '5.4.4'
```
seut, line 36:
```suggestion
            cratedb-version: '5.4.5'
```

seut (approved):
Use `5.4.5` instead?

matriv:
If 5.5.0 becomes stable today/tomorrow, maybe even wait for that one?

.github/workflows/main.yml
amotl: Is it only on macOS?
```
$ docker pull crate:5.4.5
Error response from daemon: manifest for crate:5.4.5 not found: manifest unknown: manifest unknown
```

amotl: Well that tarball link works, so it will also work here to bump to 5.4.5. Thanks.

https://cdn.crate.io/downloads/releases/crate-5.4.5.tar.gz

matriv: https://github.com/docker-library/official-images/pull/15666 is still open

Issue #402, opened 2023-10-09
Use CrateDB test layer based on "Testcontainers", in order to improve platform coverage


amotl:
## Goal

Extend platform / operating system coverage when testing CrateDB and associated infrastructure.

## Introduction

At [Testcontainers for Java and CrateDB](https://github.com/crate/cratedb-examples/tree/main/testing/testcontainers/java), we demonstrate how to run integration tests of Java applications with CrateDB, based on [Testcontainers for Java](https://github.com/testcontainers/testcontainers-java).

Based on observations at GH-343, https://github.com/crate/crate-python/pull/388, or https://github.com/mfussenegger/cr8/pull/337, we experience problems with different hand-made test layers for CrateDB, written in Python. They may have their advantages, but are not platform agnostic, so they can't be used on macOS and Windows platforms without further ado.

Currently, the integration tests here are only being invoked on Linux.
https://github.com/crate/crash/blob/623ffb33bea81fcdae38e7b0c2929ac9bc7da503/tests/test_integration.py#L28-L29

## Proposal

On behalf of two other projects, we started the corresponding [Testcontainers for Python](https://github.com/testcontainers/testcontainers-python) implementation, and converged it into the [cratedb-toolkit](https://github.com/crate-workbench/cratedb-toolkit) package, to make it reusable by downstream packages like `crash`, `crate-python`, and others.

- https://github.com/crate-workbench/cratedb-toolkit/issues/57

This issue outlines what is needed to make the test layer usable here.

- https://github.com/crate-workbench/cratedb-toolkit/issues/58

## Task

Use the `unittest`-based adapter conceived to resolve https://github.com/crate-workbench/cratedb-toolkit/issues/58 here, instead of `crate.testing.layer` in crash's [tests/test_integration.py](https://github.com/crate/crash/blob/master/tests/test_integration.py). Contributions are very welcome.

amotl:
I think this has been implemented well by @pilosus with GH-408, so we could validate and possibly close this ticket.

PR #408, opened 2023-11-21
Use testcontainers to improve platform coverage

pilosus:
## About

Use "Testcontainers" as framework for providing service fixtures.

## References

- GH-402
- https://github.com/crate-workbench/cratedb-toolkit/pull/82

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,8 +78,7 @@ setup(
         test=[
             'crate[test]',
             'zc.customdoctests<2',
-            # FIXME once tested and merged, fix the name
-            'cratedb-toolkit[test] @ git+https://github.com/crate-workbench/cratedb-toolkit@main',
+            'cratedb-toolkit[test]',
         ],
         devel=[
             'coverage<8',
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,7 @@ setup(
             'crate[test]',
             'zc.customdoctests<2',
             # FIXME once tested and merged, fix the name
-            'cratedb-toolkit[test] @ git+https://github.com/pilosus/cratedb-toolkit@tech/18-adapt-testcontainers-to-unittest',
+            'cratedb-toolkit[test] @ git+https://github.com/crate-workbench/cratedb-toolkit@main',
         ],
         devel=[
             'coverage<8',
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,9 @@ setup(
     extras_require=dict(
         test=[
             'crate[test]',
-            'zc.customdoctests<2'
+            'zc.customdoctests<2',
+            # FIXME once tested and merged, fix the name
+            'cratedb-toolkit[test] @ git+https://github.com/pilosus/cratedb-toolkit@tech/18-adapt-testcontainers-to-unittest',
         ],
         devel=[
             'coverage<8',
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -223,7 +223,7 @@ class CommandTest(TestCase):
                               "| name | name |",
                               "+------+------+",
                               "+------+------+\n"])
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint([], ['name', 'name'])
                 self.assertEqual(expected, output.getvalue())
@@ -235,7 +235,7 @@ class CommandTest(TestCase):
                               "+---------+",
                               "|    0.50 |",
                               "+---------+\n"])
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint([["0.50"]], ['version'])
                 self.assertEqual(expected, output.getvalue())
@@ -384,7 +384,7 @@ class CommandTest(TestCase):
                               '|    1 |',
                               '| NULL |',
                               '+------+\n'])
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['x'])
                 self.assertEqual(expected, output.getvalue())
@@ -400,7 +400,7 @@ class CommandTest(TestCase):
                               '| FALSE |',
                               '| 1     |',
                               '+-------+\n'])
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['x'])
                 self.assertEqual(expected, output.getvalue())
@@ -417,7 +417,7 @@ class CommandTest(TestCase):
                               '| FALSE |',
                               '| 1     |',
                               '+-------+\n'])
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['x\ny'])
                 self.assertEqual(expected, output.getvalue())
@@ -436,7 +436,7 @@ class CommandTest(TestCase):
                               '|   name string         |     |   |',
                               '| )                     |     |   |',
                               '+-----------------------+-----+---+\n'])
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['show create table foo', 'a', 'b'])
                 self.assertEqual(expected, output.getvalue())
@@ -458,7 +458,7 @@ class CommandTest(TestCase):
                               '|                                    | Planet      |',
                               '+------------------------------------+-------------+\n'])
 
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['min(name)', 'kind'])
                 # assert 0
@@ -481,7 +481,7 @@ class CommandTest(TestCase):
                               '| Galactic Sector QQ7 Active J Gamma | Galaxy      |',
                               '+------------------------------------+-------------+\n'])
 
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['min(name)', 'kind'])
                 self.assertEqual(expected, output.getvalue())
@@ -505,7 +505,7 @@ class CommandTest(TestCase):
                               '| Alpha Centauri      | Alpha - Centauri      |',
                               '+---------------------+-----------------------+\n'])
 
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['name', 'replaced'])
                 self.assertEqual(expected, output.getvalue())
@@ -527,7 +527,7 @@ class CommandTest(TestCase):
                               '| Features and conformance views | FALSE        | 3              | SQL_LANGUAGES view |',
                               '+--------------------------------+--------------+----------------+--------------------+\n'])
 
-        with CrateShell() as cmd:
+        with CrateShell(crate_hosts=[node.http_url]) as cmd:
             with patch('sys.stdout', new_callable=StringIO) as output:
                 cmd.pprint(rows, cols=['feature_name', 'is_supported', 'sub_feature_id', 'sub_feature_name'])
                 self.assertEqual(expected, output.getvalue())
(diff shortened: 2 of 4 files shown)


amotl (commented):
Dear Vitaly,

this looks looks excellent, thank you very much.

CI says something about this:
```
ERROR: Could not find a version that satisfies the requirement click-aliases<2,>=1.0.3 (from cratedb-toolkit[test]) (from versions: 1.0.0, 1.0.1)
ERROR: No matching distribution found for click-aliases<2,>=1.0.3
```
-- https://github.com/crate/crash/actions/runs/6950117802/job/18922855944#step:4:124

... but that is totally not your fault. We recently needed to add https://github.com/crate-workbench/cratedb-toolkit/commit/0a16a101cb45, and that will probably need some rebalancing we will take care about on behalf of the next iteration.

With kind regards,
Andreas.

pilosus:
@amotl yes, this is because of the Python version constraints in `click-aliases`. But the `cratedb-toolkit` itself has the same constraint of `>=3.8`, we simply don't see it, as the `pip install` fails before all the transitive deps are installed.
Tbh, I think Python 3.7 constrained can be lifted to `3.8` safely:
- Python 3.7 are almost non-existen in the `crash` [download stats](https://pypistats.org/packages/crash)
- Python 3.7 has reached the [end of life](https://devguide.python.org/versions/) this summer

amotl:
> Python 3.7 has reached the [end of life](https://devguide.python.org/versions/) this summer

urllib3 also just removed support, see https://github.com/urllib3/urllib3/releases/tag/2.1.0.

setup.py
```diff
@@ -76,7 +76,9 @@ def read(path):
     extras_require=dict(
         test=[
             'crate[test]',
-            'zc.customdoctests<2'
+            'zc.customdoctests<2',
+            # FIXME once tested and merged, fix the name
```
pilosus, line 80: NB! Needs to be updated before the merge

pilosus:
@amotl  I think I am ready for the review, please let me know if I missed something.

tests/test_integration.py
```diff
@@ -22,32 +25,34 @@
...
+        "cluster.name": "Testing44209",
+        "node.name": "crate",
+        "lang.js.enabled": True,
+        "psql.port": psql_port,
+        "http.port": http_port,
+        "transport.tcp.port": transport_port,
+    }
 
 
+node = CrateDBFixture(crate_version=EntrypointOpts.version)
```
amotl, line 46: With respect to your question on the other patch, could this spot also be used to optionally register table names for resetting, on behalf of a second argument to the fixture manager here? In this way, it can be managed by `CrateDBFixture`, but the user has control about which tables need to be defined here.

The list of tables could be supplied to the constructor, or optionally by using a dedicated method, or both. It is just an idea, I don't have any strong opinions about the "how".
```python
node.register_tables_for_reset(*tablenames: t.List[str])
```

amotl: You already responded to this at https://github.com/crate-workbench/cratedb-toolkit/pull/82#discussion_r1402805521.

Your proposed interface is perfectly lean.
```python
cratedb_service.reset(tables=RESET_TABLES)
```

pilosus: Yes, I think passing the table list to the `reset` method is even better than using it in the class constructor/class method, as you can adjust the list with more flexibility with no need to change the main fixture.

setup.py
amotl: Thank you. See also https://github.com/crate/crash/pull/410#issuecomment-1824711962:

> 3. @pilosus: When following the (in my opinion) right strategy outlined at 2., I am humbly asking for your patience on this. It would need to wait a bit for the testcontainers code to be upstreamed beforehand, [...]

\[ on which you have worked on behalf of https://github.com/crate-workbench/cratedb-toolkit/pull/82 \].

tests/test_integration.py
```diff
@@ -92,13 +96,15 @@ def test_connect(self):
...
+    def setUp(self):
+        node.reset()
 
     def _output_format(self, format, func, query="select name from sys.cluster"):
         orig_argv = sys.argv[:]
         try:
             sys.argv = ["testcrash",
                         "-c", query,
-                        "--hosts", node.http_url,
+                        "--hosts", node.get_http_url(),
```
amotl, line 107: Ah, now I see the reason why you introduced `http_url` as a _property_ to the framework-agnostic test adapter. Apologies for not following about the rationale right away.

Now, as we can see the consequences this would have to a downstream code base, my suggestion would be to re-add the `http_url` property there for compatibility reasons. I think it will do no harm to have it, additionally to the canonical `get_http_url()` method.

amotl (changes requested):

tests/test_integration.py
```diff
@@ -286,7 +292,7 @@ def test_multiple_hosts(self):
                     output = output.getvalue()
                     lines = output.split('\n')
                     self.assertRegex(lines[3], r'^\| http://[\d\.:]+ .*\| NULL .*\| FALSE .*\| Server not available')
-                    self.assertRegex(lines[4], r'^\| http://[\d\.:]+. *\| crate .*\| TRUE .*\| OK')
+                    self.assertRegex(lines[4], r'^\| http://.*@?localhost:\d+ *\| crate .*\| TRUE .*\| OK')
```
matriv, line 295: Is this always `localhost` because of testcontainers now?

pilosus: it really depends on two things:
1. if you pass `host` explicitly or rely on the default `None` value, when invoking `get_connection_url` method of the `CrateDBTestAdapter`
2. where the code is running, on the host or Docker-in-Docker, see the logic [here](https://github.com/testcontainers/testcontainers-python/blob/v3.7.1/testcontainers/core/container.py#L86). In our case, both for local development and in the CI pipelines, it will always be `localhost`

so, I'd say, it's safe to assume `localhost` won't break anything, unless `testcontainers` upstream or CI/CD pipelines in `crash` repo will be changed

tests/test_integration.py
pilosus: Fixed

tests/test_integration.py
matriv: I see, thx!

matriv (approved):
Thank you! Please wait for other reviewers though.

tests/test_integration.py
amotl: Thank you!

seut (changes requested):
Nice work and thank you a lot for your efforts @pilosus! 
But related to the discussion we had about adding dependencies to incubating projects and the future of the testcontainers, please do not merge this until the testcontainer is merged upstream.
See https://github.com/crate/crash/pull/410#issuecomment-1824711962.

amotl:
Hi.

I just invoked the test harness, and while it succeeds:
```
python -m unittest -v
Ran 103 tests in 8.408s
```

.... it clutters the log with such messages:
```
2024-01-19 21:30:31,578 [urllib3.connectionpool              ] WARNING : Retrying (Retry(total=8, connect=None, read=0, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x113ba4d50>: Failed to establish a new connection: [Errno 61] Connection refused')': /
2024-01-19 21:30:31,579 [urllib3.connectionpool              ] WARNING : Retrying (Retry(total=7, connect=None, read=0, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x113b4f590>: Failed to establish a new connection: [Errno 61] Connection refused')': /
2024-01-19 21:30:31,580 [urllib3.connectionpool              ] WARNING : Retrying (Retry(total=6, connect=None, read=0, redirect=None, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x113b3eed0>: Failed to establish a new connection: [Errno 61] Connection refused')': /
```

Do you know where this might be coming from, @pilosus?

With kind regards,
Andreas.

setup.py
amotl: I've adjusted the dependencies using 3191babfa6a.

pilosus:
> .... it clutters the log with such messages:

Hi @amotl 

My educated guess is that since the PR was open, a new major version of `docker` was [released](https://github.com/docker/docker-py/releases/tag/7.0.0). It has `requests` package bumped to a new version, which in turns uses `urllib3` of version 2 distributed rather than vendored as it used to be in earlier versions. Here's a longer discussion:
https://github.com/docker/docker-py/issues/3113

Since `testcontainers` use `docker` package with no pinned upper bound, the behaviour from `requests` has changed in `crash` too.

From what I see, it's not only `testcontainers`-related tests that are emitting these annoying warnings, but some others, like `tests/test_commands.py` since they are instantiating `crate.crash.command.CrateShell` which is invoking `crate.crash.command.CrateShell._connect`.

I think it's better to use either:

```
urllib3.disable_warnings()
```

or a module-wide fixtures like:

```python
def setUpModule():
    import logging
    logging.getLogger("urllib3").propagate = False
```

> Python 3.12, CrateDB 5.5.0 on macos-latest

Not sure what's happening here, but again, probably it has something to do with the new docker-py.

pilosus:
@amotl reg. MacOS tests failing. It's because Docker isn't pre-installed on the GitHub-runners!
https://github.com/orgs/community/discussions/25777

If you open a job, then click on `Set up job` -> `Runner Image` -> `Included Software`, you'll see the link to a list of preinstalled runner software.
Compare [ubuntu](https://github.com/actions/runner-images/blob/ubuntu22/20240116.3/images/ubuntu/Ubuntu2204-Readme.md) with [macos-12](https://github.com/actions/runner-images/blob/macOS-12/20231216.1/images/macos/macos-12-Readme.md), docker is not installed because of the licensing issues.

So, `testcontainers` won't work on MacOS runner, unfortunately :-/

amotl:
@pilosus:
> reg. MacOS tests failing. It's because Docker isn't pre-installed on the GitHub-runners!

Yeah right, that's why we skip integration tests on all other OS than Linux.

https://github.com/crate/crash/blob/78cf68258fb1425844c6ec4565e5deac0a0e36c8/tests/test_integration.py#L28-L29

Apparently, this would need to (additionally?) go into `setUpModule()` now?

https://github.com/crate/crash/blob/78cf68258fb1425844c6ec4565e5deac0a0e36c8/tests/test_integration.py#L49-L50

Otherwise, [it tries to dial the Docker API](https://github.com/crate/crash/actions/runs/7589617018/job/20674843600?pr=408#step:5:367), which goes south.

amotl:
>> .... it clutters the log with such messages:
>
> My educated guess is that since the PR was open, a new major version of `docker` was [released](https://github.com/docker/docker-py/releases/tag/7.0.0). It has `requests` package bumped to a new version, which in turns uses `urllib3` of version 2 distributed rather than vendored as it used to be in earlier versions. Here's a longer discussion: [...]

Oh wow, thanks for digging into this. I hadn't expected this. Instead, I was expecting that the CrateDB HTTP driver might have flaw, and is responsible for the log flodding coming from urllib3. Without reading https://github.com/docker/docker-py/issues/3113 yet, can you rule out that the HTTP driver needs to be investigated, and that it's coming from Docker?

If so, I will be happy to disable corresponding warnings as you suggested. However, it would be nicer if this could only be masked for requests to the Docker API, if that is the culprit here, in order not to obfuscate any other warnings originating at urllib3. Do you see a chance to achieve this?

pilosus:
> Yeah right, thanks for the reminder. That's also why the code skip integration tests on all other OS than Linux, so it will fortunately not be a regression.
> 
> https://github.com/crate/crash/blob/78cf68258fb1425844c6ec4565e5deac0a0e36c8/tests/test_integration.py#L28-L29
> 
> Apparently, the `raise SkipTest(...)` would need to (additionally?) go into `setUpModule()` now?

@amotl Check out how I've impemented it in d83177f692f78583d529514e2b62fc06f6807550

pilosus:
> Oh wow, thanks for digging into this. I hadn't expected that. Instead, I was expecting that the CrateDB HTTP driver might have a flaw, and is responsible for the log flodding coming from urllib3. Without reading [docker/docker-py#3113](https://github.com/docker/docker-py/issues/3113) yet, can you rule out that the HTTP driver needs to be investigated, and that it's definitively coming from Docker?
> 
> If so, I will be happy to disable corresponding warnings as you suggested. However, it would be nicer if this could only be masked for requests to the Docker API, if that is the culprit here, in order not to obfuscate any other warnings originating at urllib3. Do you see a chance to achieve this?

@amotl hmm, all this flood of `urllib3` warning logs are coming from the [CrateShell](https://github.com/crate/crash/blob/df8fd3582d827c788aa8f4ad69f2ab442b9a8784/crate/crash/command.py#L323) when we forgot pass in proper hosts so that it falls back to the default `localhost:4200`. I've fixed that in 58b0aef05a33417832dae0e3ed85a1b5b0f82379, except (obviously) for the case where we check multi hosts connection with one URL not available.

There's the same problem in some other, non-integration tests like `tests.test_commands.ShowTablesCommandTest.test_pre_0_57` and `tests.test_commands.MultipleStatementsTest.test_single_sql_statement_multiple_lines` just to name a few. It seems we don't mock them properly. I haven't fixed those, I guess it's better to do in a separate PR.

setup.py
amotl: We observe that the test adapter may need a few more improvements, so we postponed the plan about the upstreaming. We released the most recent updates with [cratedb-toolkit](https://pypi.org/project/cratedb-toolkit/) 0.0.4. Can I ask you to upgrade?

amotl:
Thank you very much for improving the test skipping on unsupported OS, and for tracking down the reason for the flood of urllib3 warnings, and fixing it. Excellent job! 💯

setup.py
pilosus: Done

pilosus:
Hi @amotl let me know if there's anything else I need to do to get this PR merged? 
If there's a blocker may be it's worth to mention it here as well :-)

amotl:
Thanks @pilosus. I think it will be good to go. We will have a final conversation with @seut about it, and then probably just merge, when no other objections come up. Thanks again for your efforts!

amotl (approved):

pilosus:
> I think some commits would makes sense to be squashed, but will leave that up to you.

@amotl @seut I've squashed 6 commits. Feel free to squash even more if needed.

Annnnnnnnki (approved):

PR #410, opened 2023-11-22
Remove QA support for Python 3.7

amotl:
## About

As suggested by @pilosus at GH-408, this patch drops official support for Python 3.7.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
         cratedb-version: ['4.8.4', '5.4.5']
         include:
           - os: 'macos-latest'
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,7 @@ Changes for crash
 Unreleased
 ==========
 
+- Remove QA support for Python 3.7. Thanks, @pilosus.
 
 2023/07/06 0.30.0
 =================
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,6 @@ setup(
         'Operating System :: OS Independent',
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',


matriv (approved):
I saw just a few downloads with 3.7, if you think it's fine to drop, I also don't have any objection.

seut (changes requested):
If we remove QA for 3.7, we should remove general 3.7 support. Otherwise we don't know if 3.7 is still working.  Or what is the rational about this otherwise?
We should also update the relevant documentation which seems to be quite outdated anyhow as 3.4 is still mentioned as supported. https://cratedb.com/docs/crate/crash/en/latest/getting-started.html

pilosus:
Basically, Python 3.8 or higher is needed for **testing/developing only**, as `cratedb-toolkit` itself and at least one of its transitive deps (`click-aliases`) require 3.8+. Given that we want to merge some changes to `crash` that require `cratedb-toolkit` for `test` extra, there are few options:
1. Drop 3.7 support in `crash` in favour of 3.8, see my motivation/stats [here](https://github.com/crate/crash/pull/408#issuecomment-1822576219)
2. (this PR) Provide 3.7 general support in `crash`, but 3.8+ for testing/development. Pros: easy, Cons: inconsistent, error proned, given CI's testing matrix must be adjusted
3. Fix `cratedb-toolkit` to support 3.7: get rid/fork/write our own alternative to `click-aliases` (it's just a single class under 100 LoC). Pros: Python version support is consitent across the projects, Cons: there might be more than 1 Python 3.8+ third-party deps in `cratedb-toolkit`, need to research first

pilosus:
> cratedb-toolkit seems to have an extremely broad scope, and mentions it's alpha/beta/incubation-quality code. Are we sure we want to include that here?

my two cents:
`cratedb-toolkit` is meant, among other things, to contain `testcontainers` and all the wrappers/fixtures around it.
We are going to use them in `crash` as well, see #402

pilosus:
Btw, reg. this one:
> Fix cratedb-toolkit to support 3.7: get rid/fork/write our own alternative to click-aliases
I've compiled both 3.7 and 3.8 and it seems that `click-aliases` is the only culprit that prevents using Python 3.7. 
The code of `click-aliases` will be working just fine on 3.7. It's worth submitting the PR there, I guess

pilosus:
Ah, @amotl was quick! 
https://github.com/click-contrib/click-aliases/pull/14

amotl:
Hi there,

thanks for your excellent feedback on this topic. I've also talked to @seut about it, and we decided not to bring in this patch for now, together with those additional thoughts, also responding to your requests and questions:

1. While Python 3.7 is end-of-life, there is currently no strict need to drop support for it from the perspective of `crash`.
2. While the Testcontainer stuff has been incubating in cratedb-toolkit, it does not need to stay like this. The plan is actually to upstream it to https://github.com/testcontainers/testcontainers-python. I think it might be the right time now to already start working on this, after https://github.com/crate-workbench/cratedb-toolkit/pull/82 is properly integrated and absorbed.
3. @pilosus: When following the (in my opinion) right strategy outlined at 2., I am humbly asking for your patience regarding GH-408. It would need to wait a bit for the testcontainers code to be upstreamed beforehand.

With kind regards,
Andreas.

PR #413, opened 2023-11-24
CI: Turn off failing fast on the test matrix slots

amotl:
## About

This is meant to support GH-408, so that we can get successful outcomes on the Python >= 3.8 slots already.

It is intended to be a temporary change, for a better overview. It can be reverted to save resources on daily operations later.


seut (changes requested):
Why not add this change directly to the relevant PR #408 instead?

amotl:
> Why not add this change directly to the relevant PR https://github.com/crate/crash/pull/408 instead?

I did not want to push to other's branches without previous coordination. But now that we are talking about it, we can well do it like this. Alternatively, @pilosus might just want to pick it up without further ado? ;]

PR #418, opened 2023-12-20
Add support for external pagers via \pager command

mfussenegger:
Allows to use apps like `jless` or `pspg` to view the result sets.

See:

recording


amotl (approved):
Thank you very much. 

While it doesn't need to be part of this patch, I think we should also add corresponding documentation to educate users about this feature.

On this guideline, I think roughly dumping the ASCII content of your screencast in the PR description into a corresponding reStructuredText section would be perfectly fine.

Issue #370, opened 2022-06-27
Attempting to connect to wrong port (PG post), hangs


matriv:
```
crash --host "localhost:5433" --verbose  
```
Runs for long time, without returning any error.

amotl:
Hi Marios,

thank you for reporting this. I can confirm it by using, for example:
```
nc -l -p 5433
crash --host "localhost:5433" --verbose
```

Because `crash` uses `timeout=None`, the default timeout value for "connect" and "read" will be used, which is "infinite". The `requests` library behaves equally. There is some discussion around the topic at https://github.com/urllib3/urllib3/issues/655.

In different spots, I observed others to use default timeout values like those:
```python
from urllib3.util.timeout import Timeout
Timeout(connect=2, read=7)
Timeout(connect=3.05, read=27)
Timeout(connect=5.0, read=?)
Timeout(connect=10.0, read=?)
```

Do you think it is a good idea to also set reasonable non-infinite default timeout values for `crash`? If so, what do you think which values we should use?

With kind regards,
Andreas.

matriv:
I think yes we should. It seems to me that a user would expect to see and error message if connection cannot be established.
I would say that 5 or even 10 seconds is a reasonable value, and probably it would be nice to be able to configure it on demand.

mfussenegger:
+1 for connect timeout. E.g. 5sec by default
-1 for read timeout.

matriv:
Could we also have some "keep-alive" test that can happen in the background, so even if there is a long running query, we can still get an error that the server(s) is/are disconnected?

proddata:
> Otherwise, do you think we should also configure the read timeout to a maximum value by default? If yes, what do think would be an optimal default value? 24 hours?
I think there should be none. 24h might be even worse

matriv:
> Could we also have some "keep-alive" test that can happen in the background, so even if there is a long running query, we can still get an error that the server(s) is/are disconnected?

This would address the issue that a client issues a long-run query and the hosts it's connected go down, but the client still "thinks" the query is still running. WDYT?

amotl:
Dear Marios,

sure we can think about such an "availability background check" feature. However, let's track it on behalf of a different issue instead? I think the scope of this one would be to only adjust the connect timeout accordingly. I also think five seconds as a default value, with an option to adjust it, would be fine.

With kind regards,
Andreas.

matriv:
[...]

PR #421, opened 2024-01-12
Add command-line argument `--timeout`, to configure network timeout values

amotl:
## About

Coming from GH-370, this patch introduces the `--timeout` command-line argument to specify the (connect, read) timeout values. The default connect timeout is five seconds now, the default read timeout is infinite.

## Backlog
- Needs https://github.com/crate/crate-python/pull/603.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,7 +4,10 @@ Changes for crash
 
 Unreleased
 ==========
-
+- Added command-line argument ``--timeout``, to configure network timeout
+  values in seconds. The default connect timeout is five seconds now,
+  the default read timeout is the default setting of the ``socket`` module,
+  which is "infinite" by default.
 
 2024/01/12 0.30.2
 =================
--- a/crate/crash/command.py
+++ b/crate/crash/command.py
@@ -139,6 +139,9 @@ def get_parser(output_formats=[], conf=None):
     parser.add_argument('--hosts', type=str, nargs='*',
                         default=_conf_or_default('hosts', ['localhost:4200']),
                         help='connect to HOSTS.', metavar='HOSTS')
+    parser.add_argument('--timeout', type=str, metavar='TIMEOUT',
+                        help='Configure network timeout in "<connect_sec>" or "<connect_sec>,<read_sec>" format. '
+                             'Defaults to "5,-1" where "-1" means infinite timeout.', default="5")
     parser.add_argument(
         '--verify-ssl',
         choices=(True, False),
@@ -618,8 +621,62 @@ def main():
     save_and_exit()
 
 
+INFINITE_TIMEOUT = -1
+INFINITE_TIMEOUTS = [None, INFINITE_TIMEOUT, str(INFINITE_TIMEOUT)]
+
+
+def _decode_timeout(value):
+    """
+    Decode single timeout value, respecting the `INFINITE_TIMEOUT` surrogate.
+    """
+    if value in INFINITE_TIMEOUTS:
+        return None
+    else:
+        return float(value)
+
+
+def _decode_timeouts(timeout):
+    """
+    Decode connect/read timeout values from tuple or string.
+
+    Variant 1: (connect, read)
+    Variant 2: connect,read
+    """
+
+    if timeout is None:
+        timeouts = (INFINITE_TIMEOUT, INFINITE_TIMEOUT)
+    elif isinstance(timeout, (int, float)):
+        timeouts = (timeout, INFINITE_TIMEOUT)
+    elif isinstance(timeout, str):
+        timeouts = timeout.split(",")
+    elif isinstance(timeout, (list, tuple)):
+        timeouts = timeout
+    else:
+        raise TypeError(f"Cannot decode timeout value from type `{type(timeout)}`, "
+                        f"expected format `<connect_sec>,<read_sec>`")
+
+    if len(timeouts) == 1:
+        connect_timeout, read_timeout = timeouts[0], INFINITE_TIMEOUT
+    elif len(timeouts) == 2:
+        connect_timeout, read_timeout = timeouts[0], timeouts[1]
+    else:
+        raise ValueError(f"Cannot decode timeout `{timeout}`, "
+                         f"expected format `<connect_sec>,<read_sec>`")
+
+    return urllib3.Timeout(connect=_decode_timeout(connect_timeout), read=_decode_timeout(read_timeout))
+
+
 def _create_shell(crate_hosts, error_trace, output_writer, is_tty, args,
                   timeout=None, password=None):
+
+    # Explicit "timeout" function argument takes precedence.
+    if timeout is not None:
+        timeout = _decode_timeouts(timeout)
+
+    # Probe `--timeout`` command line argument second.
+    elif args.timeout is not None:
+        timeout = _decode_timeouts(args.timeout)
+
     return CrateShell(crate_hosts,
                       error_trace=error_trace,
                       output_writer=output_writer,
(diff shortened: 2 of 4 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,10 @@ Changes for crash
 Unreleased
 ==========
 
+- Added command-line argument ``--timeout``, to configure network timeout
+  values in seconds. The default connect timeout is five seconds now,
+  the default read timeout is infinite.
```
amotl, line 10: It is not strictly "infinite", but actually uses the system default as provided by Python's `socket` library. By default, it uses `None`, which means infinite. However, it can be changed, so this detail might want to be reflected here. WDYT?

amotl: Done.

crate/crash/command.py
```diff
@@ -139,6 +139,10 @@ def _conf_or_default(key, value):
     parser.add_argument('--hosts', type=str, nargs='*',
                         default=_conf_or_default('hosts', ['localhost:4200']),
                         help='connect to HOSTS.', metavar='HOSTS')
+    parser.add_argument('--timeout', type=str, metavar='TIMEOUT',
+                        help='Configure "connect,read" network timeout values different than "5,infinite" seconds.'
+                             'Obtains either a scalar integer or float value used as connect timeout,'
+                             'or a tuple of connect- vs. read-timeout, separated by a comma.', default="5")
```
matriv, line 145:
```suggestion
                             'or a tuple of connect- and read-timeout, separated by a comma.', default="5")
```

crate/crash/command.py
```diff
@@ -139,6 +139,10 @@ def _conf_or_default(key, value):
     parser.add_argument('--hosts', type=str, nargs='*',
                         default=_conf_or_default('hosts', ['localhost:4200']),
                         help='connect to HOSTS.', metavar='HOSTS')
+    parser.add_argument('--timeout', type=str, metavar='TIMEOUT',
+                        help='Configure "connect,read" network timeout values different than "5,infinite" seconds.'
```
matriv, line 143: just double checking, `infinite` is a valid value that is parsed by `urllib3.Timout()`, right?

CHANGES.txt
```diff
@@ -4,7 +4,10 @@ Changes for crash
 
 Unreleased
 ==========
-
+- Added command-line argument ``--timeout``, to configure network timeout
+  values in seconds. The default connect timeout is five seconds now,
+  the default read timeout is the default setting of the ``socket`` module,
+  which in turn is "infinite" by default.
```
matriv, line 10:
```suggestion
  which is "infinite" by default.
```

crate/crash/command.py
```diff
@@ -620,6 +624,23 @@ def save_and_exit():
 
 def _create_shell(crate_hosts, error_trace, output_writer, is_tty, args,
                   timeout=None, password=None):
+
+    # Explicit "timeout" function argument takes precedence.
```
matriv, line 628: didn't dive into it, but where can `timeout` come from if not from the args?

matriv (commented):
thx, left some comments/questions

crate/crash/command.py
amotl: In this case, it can come from `args.timeout`, and/or from the dedicated keyword argument `timeout`.

crate/crash/command.py
amotl: No, `"infinite"` isn't a valid value, it is just a surrogate. When obtaining Python's `None` symbol, the timeout will be infinite. Currently, there is no way to obtain this value from a text representation from the command line.

amotl: Do you have a suggestion how to improve code or docs around this topic?

surister (approved):

crate/crash/command.py
amotl: Yeah. In this case, I think it would be more about discussing the need to find a textual representation. Would you prefer that the literal `"None"` would be translated to `None`, or would you prefer that `inf` or `infinite` would be translated to `None`?

Without implementing that, I think there would currently be no possibility to configure, for example, an infinite connect timeout. We would need to check / validate what `--timeout=,42` does, and eventually shape it a bit more to converge this merely undefined behaviour into something with a more sensible semantic.

amotl: Currently, omitting one or both values fails like:
```
$ crash --timeout=42,
$ crash --timeout=,42
$ crash --timeout=,
could not convert string to float: ''
```

So, momentarily, only those invokations work, adjusting either the connect timeout value, or both.
```
$ crash --timeout=42
$ crash --timeout=42,42
```

Maybe it is not needed to improve this further? Maybe it would be needed to improve error handling and presentation to the user? WDYT?

matriv: imho, we should give the chance to the user to configure the default "infinite" value, if needed,
so I'd go with `None` (case insensitive though, so that none, NONE, NoNE, etc. is all the same)

crate/crash/command.py
matriv: dedicated keyword timeout ? what do you mean, can you give an example from user's perspective?

amotl: Hi. I can't provide a user example, but I can better outline the two possible ways of obtaining the timeout value in this case.

- One way is `args.timeout`, which carries the string value from the command line option `--timeout`.
- Further, I discovered a Python kwargs called `timeout=None`, which apparently can be used to directly obtain a Python timout value (scalar or object, no string).

image

amotl: It can always be used by the user via Python API. It does not necessarily needs to be used on the command line. Removing code will probably be a different patch though, as it will need to address backward-incompatibility concerns, at least as a changelog item.

matriv: Ok, thx!

crate/crash/command.py
```diff
@@ -139,6 +139,10 @@ def _conf_or_default(key, value):
     parser.add_argument('--hosts', type=str, nargs='*',
                         default=_conf_or_default('hosts', ['localhost:4200']),
                         help='connect to HOSTS.', metavar='HOSTS')
+    parser.add_argument('--timeout', type=str, metavar='TIMEOUT',
+                        help='Configure "connect,read" network timeout values different than "5,infinite" seconds.'
+                             'Obtains either a scalar integer or float value used as connect timeout,'
+                             'or a tuple of connect- and read-timeout, separated by a comma.', default="5")
```
mfussenegger, line 145:
```suggestion
                        help='Configure network timeout in "<connect_sec>" or "<connect_sec>,<read_sec>" format. Defaults to "5,infinite" where "infinite" means no timeout.')
```

crate/crash/command.py
mfussenegger: Good call. 0 or -1 are commonly used to set no-timeout/infinite. +1

amotl: The patch has been amended to implement your suggestion. Now, `-1` can be used as a surrogate value to represent `None`, which means "infinite timeout".

amotl: The patch has been significantly amended, so I am humbly asking for another review. Thank you again.

crate/crash/command.py
```diff
@@ -618,8 +621,60 @@ def save_and_exit():
...
+        timeouts = timeout
+    else:
+        raise TypeError(f"Decoding timeout values failed. Invalid type: {type(timeout)}")
+
+    if len(timeouts) == 1:
+        connect_timeout, read_timeout = timeouts[0], INFINITE_TIMEOUT
+    elif len(timeouts) == 2:
+        connect_timeout, read_timeout = timeouts[0], timeouts[1]
+    else:
+        raise ValueError(f"Decoding timeout values failed: {timeout}")
```
mfussenegger, line 662:
```suggestion
        raise ValueError(f"Cannot decode timeout {timeout}, expected format `<connect_sec>,<read_sec>`")
```

mfussenegger (approved):

crate/crash/command.py
amotl: Thanks. I've adjust this error message, and also aligned the other one for the `TypeError` a few lines above.

PR #431, opened 2024-03-07
Add lazyload to image

msbt:

Tested locally, should just work™


amotl (approved):
Thanks!

PR #432, opened 2024-03-16
Fix status line display of the executed SQL command

amotl:
## Problem
When prefixing an SQL statement with an SQL comment (both per-line `--` and block `/* */`), crash mistreats that comment as an SQL command when printing its status line.

image

**Full Example:**
```sql
$ echo '/*foo*/ select 1' | crash
CONNECT OK
+---+
| 1 |
+---+
| 1 |
+---+
FOO 1 row in set (0.001 sec)
```

## Solution
Because crash already uses the [sqlparse](https://pypi.org/project/sqlparse/) SQL parser, it wasn't too difficult to re-use its tokenizer already for extracting the proper canonical SQL command from the SQL expression.
```sql
$ echo '/*foo*/ select 1' | crash
CONNECT OK
+---+
| 1 |
+---+
| 1 |
+---+
SELECT 1 row in set (0.003 sec)
```

## Review FYI
The [fundamental change](https://github.com/crate/crash/pull/432#pullrequestreview-1941195847) is very slim, but the patch grew a bit in size because software tests and corresponding mocking techniques needed to be adjusted a bit for better testability, and for backward-compatibility reasons with existing function interfaces (`str` vs. `sql.Statement`).

At a few of those spots, the diff also looks a bit strange, but the changes should be sound. Please take it into consideration when reviewing. When in doubt, look at the new code, not the diff. Thanks!

## Details

`sqlparse.sql.Statement.token_first()` provides the `skip_cm` option already:

> if *skip_cm* is ``True`` (default: ``False``), comments are ignored too.

Using `statement.get_type()` would be even more DWIM:

> The returned value is a string holding an upper-cased reprint of
> the first DML or DDL keyword. If the first token in this group
> isn't a DML or DDL keyword "UNKNOWN" is returned.
>
> Whitespaces and comments at the beginning of the statement
> are ignored.

However, that would introduce a regression that the outcome of parsing non-DML/DDL keywords like `BEGIN` would yield `UNKNOWN`, so we decided against using it for now. If you think it would be the better option, please advise correspondingly.

/cc @surister, @seut, @mfussenegger

### DIFF
--- a/crate/crash/command.py
+++ b/crate/crash/command.py
@@ -26,12 +26,12 @@ from __future__ import print_function
 
 import logging
 import os
-import re
 import sys
 from argparse import ArgumentParser, ArgumentTypeError
 from collections import namedtuple
 from getpass import getpass
 from operator import itemgetter
+from typing import Union
 
 import sqlparse
 import urllib3
@@ -274,7 +274,7 @@ class CrateShell:
 
     def _process_sql(self, text):
         sql = sqlparse.format(text, strip_comments=False)
-        for statement in sqlparse.split(sql):
+        for statement in sqlparse.parse(sql):
             self._exec_and_print(statement)
 
     def exit(self):
@@ -442,7 +442,7 @@ class CrateShell:
         return False
 
     def _exec(self, statement: str) -> bool:
-        """Execute the statement, prints errors if any occurr but no results."""
+        """Execute the statement, prints errors if any, but no results."""
         try:
             self.cursor.execute(statement)
             return True
@@ -457,9 +457,10 @@ class CrateShell:
                 self.logger.critical('\n' + e.error_trace)
         return False
 
-    def _exec_and_print(self, statement: str) -> bool:
+    def _exec_and_print(self, expression: Union[str, sqlparse.sql.Statement]) -> bool:
         """Execute the statement and print the output."""
-        success = self._exec(statement)
+        statement = to_statement(expression)
+        success = self._exec(str(statement).strip())
         self.exit_code = self.exit_code or int(not success)
         if not success:
             return False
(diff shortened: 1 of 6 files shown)


crate/crash/command.py
```diff
@@ -482,9 +483,26 @@ def _exec_and_print(self, statement: str) -> bool:
         return True
 
 
-def stmt_type(statement):
+def stmt_type(expression: Union[str, sqlparse.sql.Statement]):
     """Extract type of statement, e.g. SELECT, INSERT, UPDATE, DELETE, ..."""
-    return re.findall(r'[\w]+', statement)[0].upper()
+    statement = to_statement(expression)
+    return str(statement.token_first(skip_ws=True, skip_cm=True)).upper()
```
amotl, line 489: That is the spot where the previous naive way of finding the canonical SQL command of an expression was replaced by using a more elaborate method now, completely based on sqlparse's capabilities to skip whitespace and comments.

tests/test_commands.py
```diff
@@ -236,9 +238,9 @@ def test_sql_comments(self):
 comment */ 4;
 """
         cmd = CrateShell()
-        cmd._exec_and_print = MagicMock()
+        cmd._exec = Mock()
         cmd.process_iterable(sql.splitlines())
-        self.assertListEqual(cmd._exec_and_print.mock_calls, [
+        self.assertListEqual(cmd._exec.mock_calls, [
```
amotl, line 243: That noise, and subsequent spots, have been needed to move the testing probe to a different spot. Those tests verify that what goes in (user input SQL statement) will yield the correct outcome when being submitted to the database.

Previously, the probe was sitting on `_exec_and_print` by mocking it and verifying the calls on this mock. Now, that probing moved to the more low-level `_exec` method. This changes the execution behavior of the test suite, as the `_exec_and_print` method will now be executed fully [^1], unlocking the possibility to test its effects.

[^1]: That is the reason why the test suite also needs to supply a mocked cursor now, because the ingredients of `_exec_and_print` need it.

mfussenegger (approved):
looks okay to me but I think we should look into replacing sqlparse with a parser generated from the CrateDB antlr grammar, to ensure it supports our SQL dialect. 

See https://github.com/antlr/antlr4/blob/4.6/doc/python-target.md

surister (approved):

BaurzhanSakhariev:
Not sure whether it's related but just in case https://github.com/crate/crate-alerts/issues/626

PR #434, opened 2024-03-19
Fix status line display of the executed SQL command, part 2

amotl:
After recent improvements (GH-432), SQL command _arguments_ came through verbatim:
```sql
DENY DQL, DML, DDL, AL ON SCHEMA sys TO test;
DENY DQL, DML, DDL, AL OK, 1 row affected (0.082 sec)
```

This fixes it to just display the effective SQL command, without any arguments:
```sql
DENY OK, 1 row affected (0.082 sec)
```

### DIFF
--- a/crate/crash/command.py
+++ b/crate/crash/command.py
@@ -26,6 +26,7 @@ from __future__ import print_function
 
 import logging
 import os
+import re
 import sys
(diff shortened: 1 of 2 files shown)


seut (approved):
👍

PR #435, opened 2024-03-19
Fix dependencies for testing

amotl:
## About
geojson was missing, while it should be there. No idea why that happens. geojson is a dependency of crate[sqlalchemy]. It looks like we need to be more explicit here.

## References
- https://jenkins.crate.io/job/CrateDB/job/nightly/job/crash_test_nightly/770/console

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
     },
     extras_require=dict(
         test=[
-            'crate[test]',
+            'crate[sqlalchemy,test]',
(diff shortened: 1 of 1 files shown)


BaurzhanSakhariev (approved):

matriv (approved):

BaurzhanSakhariev:
Another similar issue has popped up https://github.com/crate/crate-alerts/issues/628

PR #467, opened 2025-05-13
Update index.rst: Readable screenshot, move down GH reference

bmunkholm:

Move down Github admonition (it's secondary) and make screenshots visible in size.
Keep only last screenshot as the other two just leads up to the third.

## Preview

https://crash--467.org.readthedocs.build/en/467/

### Before

image

### After

image

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,17 +6,10 @@ The CrateDB Shell
 
 The CrateDB Shell (aka *Crash*) is an interactive `CLI`_ for working with CrateDB.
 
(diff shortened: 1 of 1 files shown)


amotl (approved):
That's an excellent layout update, thanks a stack.

RTD will also give you a preview, so you don't necessarily need to relay screenshots into the PR description: You can easily pick the URL from the pull request builder widget,

- https://crash--467.org.readthedocs.build/en/467/

and just slot this one into the text body using a dedicated section.

image

Issue #469, opened 2025-05-16
Update CI docs.yml to only build on linux


bmunkholm:
Similar to how it's done elsewhere for docs recently.

PR #470, opened 2025-05-16
CI: Stop building `docs` on `macos-latest`

amotl:
It saves a few cycles on the linkchecker, both on GHA and on remote web servers.
- GH-469


bmunkholm (commented):
Thanks for doing this!. This works perfectly fine.
If you care about it, we could optionally also get rid of the matrix for simplification, and to follow the blueprint in the other docs repos where we did this.

PR #472, opened 2025-06-16
Remove Compatibility page from Docs

bmunkholm:

There was a few leftovers of page-local TOCs. Instead of just removing those the compatibility page was removed entirely as it was mentioning unsupported versions of the client and Python. Additionally I lifted the only remaining Appendix "Response formats" out to main level.
Added references to the response formats.
 
## Preview
https://crash--472.org.readthedocs.build/en/472/


docs/formats.rst
```diff
@@ -4,12 +4,8 @@
 Response formats
 ================
 
-Crash supports multiple output formats.
-
```
amotl, line 8: Maybe keep those two lines as a minimal intro to this section?

amotl (approved):

PR #490, opened 2026-03-03
Dependencies: Permit installation of crate-python up to any 2.x.x release

amotl:
## About
crash is compatible with all versions of crate-python up until the recent 2.1.1, and will likely be compatible with higher releases of the 2.x.x series.

## Details
I've verified all those installations of crash can connect to CrateDB successfully, to support the relaxed version constraint `crate<3`.
```shell
uv pip install --upgrade crash "crate<0.10"
uv pip install --upgrade crash "crate<1"
uv pip install --upgrade crash "crate<2"
uv pip install --upgrade crash "crate<3"
```

## References
- https://github.com/crate/crate-python/issues/787

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ from setuptools import setup
 requirements = [
     'colorama<1',
     'Pygments>=2.4,<3',
-    'crate>=2.0.0.dev6',
+    'crate<3',
     'platformdirs<5',
     'prompt-toolkit>=3.0,<4',
     'tabulate>=0.9,<0.10',


matriv (approved):

setup.py
```diff
@@ -28,7 +28,7 @@
 requirements = [
     'colorama<1',
     'Pygments>=2.4,<3',
-    'crate>=2.0.0.dev6',
+    'crate>=2',
```
amotl: Also add an upper bound like the other dependencies are doing it?
```suggestion
    'crate>=2,<3',
```

amotl: Nitpick: crash is likely also compatible with crate-python 1.x.
```suggestion
    'crate<3',
```

matriv: `< 3` doesn't mean that it will use the latest version that is < 3.x.x?

matriv (approved):

setup.py
amotl: The latest version is currently 2.1.1. Yes, `<3` will mean consumers can pull all available `2.x.x` versions whatsoever, but not `3.x.x` and higher.

amotl: A version constraint like this makes sense because authors of crate will/should signal a breaking change by bumping the major release version. In this spirit, older versions of crash will be resilient about the future. I have a certain perspective here, but I am sure there are other opinions, either towards more lenient, or towards more stronger variants of version management. 🤷

matriv: I think it looks good, imnho.

Issue #493, opened 2026-03-18
Add internal command to watch shard relocations progresses


seut:
It some cases, it's very interesting to know which shards are currently relocating and also the current progress of such operations. E.g.
- in rolling restarts as shard relocations will defer the restart
- when scaling up/down
- initial shard recovery on node start up
- ...

A common query used to see shards which are currently relocating or initializing is:
```
select node['name'], id, recovery['stage'], recovery['size'], routing_state, state, primary, table_name, relocating_node, size / 1024^3 as size_gb, partition_ident
        from sys.shards
        where routing_state in ('RELOCATING', 'INITIALIZING')
        order by id;
```

At least for relocating shards, it should be possible to know the current progress. Not sure if this is feasible on initial recovery as the final sizes or maximum sequence numbers may not be known upfront.

Implementing an internal command which directly gives a user the current state without the need to know such system queries could help a lot.

mannreis:
I've started looking at this! 

Could this be a reasonable view? I 
The values on the `recovery['size']` column seems quite static so I've just picked `percentage`. And for the default query I simply inverted the `where` clause to show all but `routing_state = "STARTED"`. 
```
cr> \shards
+---------------+------------+-------------+-----------+---------------------+
| routing_state | state      | shard_count |  num_docs |             size_gb |
+---------------+------------+-------------+-----------+---------------------+
| STARTED       | STARTED    |          10 |   2000000 | 0.10163962095975876 |
| UNASSIGNED    | UNASSIGNED |           4 |         0 | 0.0                 |
| RELOCATING    | STARTED    |           1 | 100000000 | 4.6805493999272585  |
+---------------+------------+-------------+-----------+---------------------+
+-----------+----+-------+--------+---------------+------------+---------+------------+------------------------+---------+-----------------+
| name      | id | stage |      % | routing_state | state      | primary | table_name | relocating_node        | size_kb | partition_ident |
+-----------+----+-------+--------+---------------+------------+---------+------------+------------------------+---------+-----------------+
| cratedb02 |  0 | DONE  |  100.0 | RELOCATING    | STARTED    | TRUE    | big_table  | 9OZST99-RDCT9aJlAlUg3A | 4907911 |                 |
| NULL      |  0 | NULL  | NULL   | UNASSIGNED    | UNASSIGNED | FALSE   | big_table  | NULL                   |       0 |                 |
[...]

PR #494, opened 2026-03-22
Feature: add \shards command to track relocations

mannreis:

As requested in #493  this PR aims adds a new command `\shards` to conveniently query for shards undergoing relocation. It computes the the aggregated progress per table based on the sum of the sizes of relocating shards divided by the total size of the table shards.

Optionally, `state` and `relocating` can be provided as an argument to respectively:
 - get a short summary of shards aggregated by state
 - view details of which ones are currently relocating
 
## Integration testing
I resorted to `docker compose` and scripting to move shards withing a cluster in order to test the what's implemented here. In order integrate that type of testing it in this code base, I spent sometime looking at `cratedb_toolkit.testing.testcontainers.cratedb` but I believe it would require another PR. For now, I resort to single-node integration testing.

## Checklist
 - [x] Fixes https://github.com/crate/crash/issues/493
 - [x] Unit testing
 - [x] Integration testing
 - [x] Command docs
 - [x] Update `CHANGES.txt`

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,9 @@ Changes for crash
 Unreleased
 ==========
 
+- Added ``\shards`` to show shard relocation progress. With optinal arguments
+  ``state`` and ``relocating``.
+
 2026/02/09 0.32.0
 =================
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,11 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
-- Added ``\shards`` to show shard relocation progress. With optional arguments
-  ``state`` and ``relocating``.
+- Added ``\shards`` to show shard relocation progress. With optional argument ``info``.
 
-- Added ``\shards`` to show shard relocation progress. With optinal arguments
-  ``state`` and ``relocating``.
 
 2026/02/09 0.32.0
 =================
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,9 @@ Unreleased
 - Added ``\shards`` to show shard relocation progress. With optional arguments
   ``state`` and ``relocating``.
 
+- Added ``\shards`` to show shard relocation progress. With optinal arguments
+  ``state`` and ``relocating``.
+
 2026/02/09 0.32.0
 =================
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,7 +5,7 @@ Changes for crash
 Unreleased
 ==========
 
-- Added ``\shards`` to show shard relocation progress. With optinal arguments
+- Added ``\shards`` to show shard relocation progress. With optional arguments
   ``state`` and ``relocating``.
 
 2026/02/09 0.32.0
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -56,10 +56,11 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
-| ``\shards [all]``      | Queries the ``sys`` tables aggregating shards status|
-|                        | followed by filtering shards relocation and         |
-|                        | initialization, unless ``all``  is used. In which   |
-|                        | case no filter is applied.                          |
+| ``\shards [VIEW]``     | Queries ``sys.shards`` table and computes relocation|
+|                        | progress progress per table. If ``VIEW`` is instead:|
+|                        |                                                     |
+|                        | - ``state`` provides aggregeration on shard state   |
+|                        | - ``relocating`` tracks which shards are relocated  |
 +------------------------+-----------------------------------------------------+
 | ``\sysinfo``           | Query the ``sys`` tables for system and cluster     |
 |                        | information.                                        |
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -56,6 +56,11 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
+| ``\shards [all]``      | Queries the ``sys`` tables aggregating shards status|
+|                        | followed by filtering shards relocation and         |
+|                        | initialization, unless ``all``  is used. In which   |
+|                        | case no filter is applied.                          |
++------------------------+-----------------------------------------------------+
 | ``\sysinfo``           | Query the ``sys`` tables for system and cluster     |
 |                        | information.                                        |
 +------------------------+-----------------------------------------------------+
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -56,10 +56,14 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
-| ``\shards [MODE]``     | Queries ``sys.shards`` table shows an overview on   |
-|                        | the state of all shards. Providing ``MODE``  shows: |
+| ``\shards [MODE]``     | Shows the cluster's shard information. Either as    |
+|                        | as global overview or per-table.                    |
 |                        |                                                     |
-|                        | - ``info`` descriminates shard relocation per table |
+|                        | ``MODE`` can be one of the following:               |
+|                        |                                                     |
+|                        | - not set (defaults to ``overview``)                |
+|                        | - ``overview`` (grouped by shard state)             |
+|                        | - ``per-table`` (grouped by table)                  |
 +------------------------+-----------------------------------------------------+
 | ``\sysinfo``           | Query the ``sys`` tables for system and cluster     |
 |                        | information.                                        |
--- a/docs/commands.rst
+++ b/docs/commands.rst
@@ -56,11 +56,10 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
-| ``\shards [VIEW]``     | Queries ``sys.shards`` table and computes relocation|
-|                        | progress progress per table. If ``VIEW`` is instead:|
+| ``\shards [MODE]``     | Queries ``sys.shards`` table shows an overview on   |
+|                        | the state of all shards. Providing ``MODE``  shows: |
 |                        |                                                     |
-|                        | - ``state`` provides aggregeration on shard state   |
-|                        | - ``relocating`` tracks which shards are relocated  |
+|                        | - ``info`` descriminates shard relocation per table |
 +------------------------+-----------------------------------------------------+
 | ``\sysinfo``           | Query the ``sys`` tables for system and cluster     |
 |                        | information.                                        |
--- a/src/crate/crash/commands.py
+++ b/src/crate/crash/commands.py
@@ -234,6 +234,51 @@ class CheckCommand(Command):
             cmd.logger.warn('No check for {}'.format(check_name))
 
 
+class ShardsCommand(Command):
+    """ short summary for shards and their relocation (use `all` as argument for longer output) """
+
+    SHORT_STMT = """SELECT routing_state, state, COUNT(*) AS shard_count, SUM(num_docs) as num_docs, SUM(size) / 1073741824.0 as size_gb
+        FROM sys.shards
+        GROUP BY routing_state, state
+        ORDER BY shard_count DESC;
+    """
+
+    RELOC_STMT = """
+        SELECT node['name'] as name, id, recovery['stage'] as stage, recovery['size']['percent'] AS "%", routing_state, state, primary, table_name, relocating_node, size / 1024 as size_kb, partition_ident
+        FROM sys.shards
+        WHERE routing_state NOT IN ('STARTED')
+        ORDER BY id, name;
+    """
+
+    ALLSHARDS_STMT = """
+        SELECT node['name'] as name, id, recovery['stage'] as stage, recovery['size']['percent'] AS "%", routing_state, state, primary, table_name, relocating_node, size / 1024 as size_kb, partition_ident
+        FROM sys.shards
+        ORDER BY id, name;
+    """
+
+    OPTIONS = ('all',)
+
+    def complete(self, cmd, text):
+        return (i for i in self.OPTIONS if i.startswith(text) or text == '')
+
+    def execute(self, cmd, stmt):
+        success = cmd._exec(stmt)
+        cmd.exit_code = cmd.exit_code or int(not success)
+        if not success:
+            cmd.logger.warn("FAILED")
+            return False
+
+        cur = cmd.cursor
+        shards = cur.fetchall()
+        if len(shards):
+            cmd.pprint(shards, [c[0] for c in cur.description])
+        return True
+
+    def __call__(self, cmd, *args, **kwargs):
+        for stmt in (self.SHORT_STMT, self.RELOC_STMT if 'all' not in args else self.ALLSHARDS_STMT):
+            self.execute(cmd, stmt)
+
+
 built_in_commands = {
     '?': HelpCommand(),
     'r': ReadFileCommand(),
@@ -243,4 +288,5 @@ built_in_commands = {
     'verbose': ToggleVerboseCommand(),
     'check': CheckCommand(),
     'pager': SetPager(),
+    'shards': ShardsCommand(),
 }
--- a/src/crate/crash/commands.py
+++ b/src/crate/crash/commands.py
@@ -253,24 +253,24 @@ class ShardsCommand(Command):
     """
 
     INFO_STMT = """
-SELECT
-    schema_name,
-    table_name,
-    partition_ident,
-    COUNT(*)
-        AS total_shards,
-    SUM(size)
-        As total_size,
-    COUNT(*) FILTER (WHERE routing_state = 'RELOCATING')
-        AS relocating_shards,
-    SUM(size) FILTER (WHERE routing_state = 'RELOCATING')
-        AS relocating_size,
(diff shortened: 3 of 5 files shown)


src/crate/crash/commands.py
```diff
@@ -234,13 +234,104 @@ def __call__(self, cmd, check_name=None, **kwargs):
...
+    SUM(CASE WHEN routing_state = 'RELOCATING' THEN 1 ELSE 0 END)
+        AS relocating_shards,
+    SUM(CASE WHEN routing_state = 'RELOCATING' THEN size ELSE 0 END)
+        AS relocating_size,
+    100.0 * SUM(CASE WHEN routing_state != 'RELOCATING' THEN size ELSE 0 END) / CAST(SUM(size) as DOUBLE)
+        AS relocated_percent
+FROM sys.shards
+WHERE routing_state != 'UNASSIGNED'
+GROUP BY table_name
+ORDER BY relocated_percent, table_name;
```
seut: I wonder if this should be rather the `relocating` statement and the `state` maybe the default one?

I think adding `schema_name` and `partition_ident` would be helpful.
Additionally, we could make use of filters on aggregations to ease the statement (also improves performance afaik). If we treat this as the `relocating` statement, I'd leave out number of docs as not really interesting (keep focus on the relevant parts).


```suggestion
SELECT
    schema_name,
    table_name,
    partition_ident,
    COUNT(*)
        AS total_shards,
    SUM(size)
        As total_size,
    COUNT(*) FILTER (WHERE routing_state = 'RELOCATING')
        AS relocating_shards,
    SUM(size) FILTER (WHERE routing_state = 'RELOCATING')
        AS relocating_size,
    100.0 * SUM(size) FILTER(WHERE routing_state != 'RELOCATING') / SUM(size)
        AS relocated_percent
FROM sys.shards
WHERE routing_state != 'UNASSIGNED'
GROUP BY schema_name, table_name, partition_ident
ORDER BY relocated_percent, schema_name, table_name, partition_ident;
```

src/crate/crash/commands.py
```diff
@@ -234,13 +234,104 @@ def __call__(self, cmd, check_name=None, **kwargs):
...
+            routing_state,
+            COUNT(*)
+                AS shard_count,
+            SUM(num_docs)
+                AS num_docs,
+            SUM(size) / 1073741824.0
+                AS size_gb
+        FROM sys.shards
+        GROUP BY routing_state
+        ORDER BY routing_state;
```
seut: As said, I'd suggest to use this as the default statement for `\shards`.
Adding the `primary` column may be helpful, especially when UNASSIGNED shards exists as unassigned replica shards may be expected while unassigned primaries are critical.

```suggestion
        SELECT
            state,
            primary,
            COUNT(*)
                AS shard_count,
            SUM(num_docs)
                AS num_docs,
            SUM(size) / 1024^3
                AS size_gb
        FROM sys.shards
        GROUP BY state, primary
        ORDER BY state, primary;
```

src/crate/crash/commands.py
```diff
@@ -234,13 +234,104 @@ def __call__(self, cmd, check_name=None, **kwargs):
...
+            state,
+            primary,
+            relocating_node,
+            size / 1024.0
+                AS size_kb,
+            partition_ident
+        FROM sys.shards
+        WHERE routing_state = 'RELOCATING'
+        ORDER BY table_name, id, node['name'];
+    """
```
seut: I think this statement won't help much as it won't give us any information on the relocation progress. I've played around a bit and currently I see no way how we can get the progress per shard as CrateDB does not expose this (the `recovery['size']` only exposes real shard recovery and not the relocation).
Thus, we can only show a overall progress, counting complete shard sizes as you did in the existing `DEFAULT_STMT` unfortunately. Seems like we need to improve something on CrateDB to have real, per shard progresses.
For now, I suggest to remove it.

tests/test_integration.py
```diff
@@ -870,4 +872,115 @@ def test_connect_info_not_available(self, is_conn_available):
             self.assertEqual(crash.connect_info.schema, None)
 
 
+class ShardsCommandEmptyDBTest(TestCase):
+    def setUp(self):
+        node.reset()
+
+    def test_shards_command(self):
```
seut: Maybe we should name this a bit more concrete to reflect whats really tested (only output format, no values).
```suggestion
    def test_shards_command_output_format(self):
```

Same for the other test methods.

seut (commented):
I've added some suggestions related to the used statements, otherwise looks fine.
I had a bit in mind that having some ongoing progress visible with some kind of progress bar, which one could get out e.g. by ctrl-c would be awesome, but also probably better to work on this as a followup, now that you've done the ground work :)

About real integration testing: Right, this would require at least 2 nodes to make it work. Utilizing the testcontainers to make this work would be really great, but agree, lets look into this as followup as well.

seut:
@mannreis any update on this, do you need any further help to continue?

mannreis:
Your suggestions were already a great help! I'll apply them early next week. They seem easy enough to have been applied already.

Regarding your original idea of an interactive progress bar. I saw that postgres cli had a `\watch` command that seemed useful and could be used for queries but I guess other commands as well.

src/crate/crash/commands.py
mannreis: I've had issues applying this commit due an error when parsing/evaluating of `1024^3` in the query. I can only reproduce it when running the tests
```
SQLParseException[line 9:29: mismatched input '^' expecting {<EOF>, ';'}]
```
```
  src/crate/crash/commands.py(297)__call__()
-> self.execute(cmd, self.DEFAULT_STMT)
 src/crate/crash/commands.py(284)execute()
-> success = cmd._exec(stmt)
  src/crate/crash/command.py(475)_exec()
-> self.cursor.execute(statement)
  .venv/lib/python3.13/site-packages/crate/client/cursor.py(57)execute()
-> self._result = self.connection.client.sql(
  .venv/lib/python3.13/site-packages/crate/client/http.py(549)sql()
-> content = self._json_request("POST", self.path, data=data)
> .venv/lib/python3.13/site-packages/crate/client/http.py(683)_json_request()
-> _raise_for_status(response)
(Pdb) p _raise_for_status(response)
*** crate.client.exceptions.ProgrammingError: SQLParseException[line 9:29: mismatched input '^' expecting {<EOF>, ';'}]
(Pdb) p self.path
'/_sql?types=true'
(Pdb) p data
b'{"stmt":"\\n        SELECT\\n            state,\\n            primary,\\n            COUNT(*)\\n                AS shard_count,\\n            SUM(num_docs)\\n                AS num_docs,\\n            SUM(size) / 1024^3\\n                AS size_gb\\n        FROM sys.shards\\n        GROUP BY state, primary\\n        ORDER BY state, primary;\\n    "}'
```

For now I add it as a literal. Any ideia why this happens?

CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for crash
 Unreleased
 ==========
 
+- Added ``\shards`` to show shard relocation progress. With optinal arguments
```
bgunebakan:
```suggestion
- Added ``\shards`` to show shard relocation progress. With optional arguments
```

docs/commands.rst
```diff
@@ -56,6 +56,12 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
+| ``\shards [VIEW]``     | Queries ``sys.shards`` table and computes relocation|
+|                        | progress progress per table. If ``VIEW`` is instead:|
+|                        |                                                     |
+|                        | - ``state`` provides aggregeration on shard state   |
+|                        | - ``relocating`` tracks which shards are relocated  |
```
bgunebakan:
```suggestion
|                        | - ``state`` provides aggregation on shard state   |
|                        | - ``relocating`` tracks which shards are relocating  |
```

tests/test_integration.py
```diff
@@ -870,4 +872,95 @@ def test_connect_info_not_available(self, is_conn_available):
...
+            with patch('sys.stdout', new_callable=StringIO) as output:
+                text = shards_(cmd)
+                self.assertEqual(None, text)
+                output_lines = output.getvalue().splitlines()
+                self.assertEqual(len(expected), len(output_lines))
+                header = lambda x: [word.strip() for word in x[1].split('|')]
+                self.assertEqual(header(expected), header(output_lines))
+
+
+    def test_shards_command_ouput_info(self):
```
bgunebakan:
```suggestion
    def test_shards_command_output_info(self):
```

bgunebakan (approved):
Thank you so much for your contribution! Your PR looks good, there is just so small typos on naming. I think it's ready to merge after fixing them.

Best,
Bilal

florinutz:
@mfussenegger can you please have another quick look at this?

docs/commands.rst
```diff
@@ -56,6 +56,12 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
+| ``\shards [VIEW]``     | Queries ``sys.shards`` table and computes relocation|
```
mfussenegger: How about calling this `subject` or `mode` instead of `view`? `view` confused me initially a bit because I thought of it referring to https://cratedb.com/docs/crate/reference/en/latest/general/ddl/views.html

src/crate/crash/commands.py
```diff
@@ -234,13 +234,84 @@ def __call__(self, cmd, check_name=None, **kwargs):
             cmd.logger.warn('No check for {}'.format(check_name))
 
 
+class ShardsCommand(Command):
+    """shows shards state, optionally per table, e.g. \\shards info"""
```
mfussenegger: The docs refer to `state` and `relocating` as options, but here it refers to `info` ?

mfussenegger (changes requested):

florinutz:
@mannreis can you please maybe do the changes and squash-merge this? I intend to create another fork and port your code there otherwise, if that is fine with you

mannreis:
> @mannreis can you please maybe do the changes and squash-merge this? I intend to create another fork and port your code there otherwise, if that is fine with you

I can do it over this week, if it's fine for you. (I am still away from the keyboard.)

Thank you for the reviews!!!

docs/commands.rst
mannreis: Should have struck me that "view" is a reserved keyword.

src/crate/crash/commands.py
mannreis: These were artifacts that lingered in the docs. Here, the docstring is correct

bgunebakan (approved):
Thank you for your contribution!

docs/commands.rst
```diff
@@ -56,6 +56,11 @@ Every command starts with a ``\`` character.
 | ``\r <FILENAME>``      | Reads statements from ``<FILENAME>`` and execute    |
 |                        | them.                                               |
 +------------------------+-----------------------------------------------------+
+| ``\shards [MODE]``     | Queries ``sys.shards`` table shows an overview on   |
+|                        | the state of all shards. Providing ``MODE``  shows: |
+|                        |                                                     |
+|                        | - ``info`` descriminates shard relocation per table |
++------------------------+-----------------------------------------------------+
```
mfussenegger: I find it a bit confusing that omitting `mode` has a different behavior than providing `mode`.

Imho it should have two options: `overview`, `per-table` (or only `table` - not sure).
With `overview` being the default used if no mode is provided.

I'd also rephrase the description a bit. "Shows the cluster's shard information. Either as overview or per-table".
That it queries `sys.shards` is sort of an implementation detail.

tests/test_integration.py
```diff
@@ -653,10 +653,12 @@ def test_help_command(self):
             '\\pager                          set an external pager. Use without argument to reset to internal paging',
             '\\q                              quit crash',
             '\\r                              read and execute statements from a file',
+            '\\shards                         shows shards state, optionally per table, e.g. \\shards info',
```
bgunebakan:
```suggestion
            '\\shards                         shows shards overview, optionally per table, e.g. \\shards info',
```

src/crate/crash/commands.py
```diff
@@ -234,13 +234,85 @@ def __call__(self, cmd, check_name=None, **kwargs):
             cmd.logger.warn('No check for {}'.format(check_name))
 
 
+class ShardsCommand(Command):
+    """shows shards overview, optionally per table, e.g. \\shards info"""
```
bgunebakan:
```suggestion
    """shows shards overview, optionally per table, e.g. \\shards per-table"""
```

tests/test_commands.py
```diff
@@ -267,6 +268,66 @@ def test_check_command_with_node_check(self, cmd):
         cmd.logger.info.assert_called_with('NODE CHECK OK')
 
 
+class ShardsCommandTest(TestCase):
```
mfussenegger, line 272: I'd tend to remove this. Given the coverage from `ShardsCommandWithContentTest` this doesn't seem to bring any value - but will cause maintenance costs.

tests/test_integration.py
```diff
@@ -870,4 +872,109 @@ def test_connect_info_not_available(self, is_conn_available):
...
+        node.reset()
+        with CrateShell(crate_hosts=[node.http_url], is_tty=False) as cmd:
+            cmd.process('CREATE TABLE test_table (id INTEGER PRIMARY KEY, data STRING ) CLUSTERED INTO 10 SHARDS WITH (number_of_replicas = 0);\n')
+
+    def test_shards_command_output_default(self):
+        expected = [
+            '+---------+---------+-------------+-----------+---------------------+',
+            '| state   | primary | shard_count |  num_docs |             size_gb |',
+            '+---------+---------+-------------+-----------+---------------------+',
+            '| STARTED | FALSEy  |          ?6 | 1x0000000 |  7. DUMMY VALUE 614 |',
```
mfussenegger, line 948: I'm a bit confused by this expected output. Where's the `DUMMY VALUE` coming from? What's with the FALSEy, ?6, 1x00000... ?

mfussenegger (changes requested):

bgunebakan (changes requested):
Hi @mannreis, thanks for sticking with this, it's very close. I'd like to finalise the PR and merge your contribution.

tests/test_commands.py
```diff
@@ -268,6 +269,67 @@ def test_check_command_with_node_check(self, cmd):
...
+
+    @patch('crate.crash.command.CrateShell')
+    def test_shards_command_per_table_none(self,cmd):
+        cmd._exec.return_value = True
+        cmd.cursor.fetchall.return_value = []
+        ShardsCommand()(cmd,"per-table")
+        cmd.logger.info.assert_not_called()
+
+
+@patch('crate.client.connection.Cursor', fake_cursor())
```
bgunebakan:
```suggestion
```

tests/test_integration.py
```diff
@@ -653,10 +653,12 @@ def test_help_command(self):
             '\\pager                          set an external pager. Use without argument to reset to internal paging',
             '\\q                              quit crash',
             '\\r                              read and execute statements from a file',
+            '\\shards                         shows shards overview, optionally per table, e.g. \\shards info',
```
bgunebakan:
```suggestion
            '\\shards                         shows shards overview, optionally per table, e.g. \\shards per-table',
```

bgunebakan:
Status: conflicts are resolved and I've pushed the conflict resolution and the docstring change to your branch already.

To be explicit about how this lands: your commits keep you as the author, and you'll be credited. I'm only finishing the review follow-ups, not claiming the work. If you'd rather push the remaining fix yourself, say so and I'll hold off; otherwise I'll wrap it up this week.

@mfussenegger your change request from July is still open. The two points you raised (dropping the mock-based ShardsCommandTest, and the placeholder DUMMY VALUE row in the integration test) are addressed. Could you take another look once CI is green? I've authored commits on this branch now, so my own approval shouldn't be the one that lands it.

Thanks.

bgunebakan (approved):

mfussenegger:
>  The two points you raised (dropping the mock-based ShardsCommandTest, and the placeholder DUMMY VALUE row in the integration test) are addressed.

Maybe I'm missing something but to me they still appear unaddressed?

tests/test_commands.py
bgunebakan: You are right, it does not add any value. I dropped it with 5a7cc767e16b209f98ee8eaa42ae2f862a5b1ee0

tests/test_integration.py
bgunebakan: expected was read only as `len(expected)` and `expected[1]` and no cell value is compared. I removed expected and add full value assertion with 5a7cc767e16b209f98ee8eaa42ae2f862a5b1ee0

bgunebakan:
> > The two points you raised (dropping the mock-based ShardsCommandTest, and the placeholder DUMMY VALUE row in the integration test) are addressed.
> 
> Maybe I'm missing something but to me they still appear unaddressed?

I reviewed changes and your comments again, fixed them.

tests/test_integration.py
```diff
@@ -870,4 +872,102 @@ def test_connect_info_not_available(self, is_conn_available):
...
+        ]
+        with CrateShell(crate_hosts=[node.http_url], is_tty=False) as cmd:
+            shards_ = cmd.commands['shards']
+            with patch('sys.stdout', new_callable=StringIO) as output:
+                text = shards_(cmd, 'per-table')
+                self.assertEqual(None, text)
+                output_lines = output.getvalue().splitlines()
+                self.assertEqual(self.EXPECTED_LINES, len(output_lines))
+                columns = [word.strip() for word in output_lines[1].strip('|').split('|')]
+                self.assertEqual(expected_columns, columns)
```
mfussenegger, line 971: The test class `ShardsCommandWithContentTest` somehow implies that this would test the actual content, instead it now asserts the column names - which are the same independent of if there are any rows or not. 

Granted it also asserts the line count, but one could break the query without these tests failing. Shouldn't it include the output too?

bgunebakan: You're right, I added INSERT statement in `setUp()` method to create data inside the table and moved `CrateShell()` parts into `_run()` method to reduce duplication and added content asserts.

mfussenegger (approved):

PR #497, opened 2026-04-28
Add URL query param support for `verify_ssl`

### LINKED ISSUE #482: Respect URL connectivity parameters
> `crash` requires an additional cli arg `--verify-ssl False`, and does not honor this as a URL parameter.

I would be cool to bring those feature to `crash` as well?

florinutz:

Adds support for `?verify_ssl=<bool>` in the host URL passed to `--hosts`. 

The cli flag `--verify-ssl` continues to take precedence over the URL param.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #482

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for crash
 Unreleased
 ==========
 
+- Honor ``?verify_ssl=<bool>`` in the host URLs to toggle TLS certificate verification. The CLI ``--verify-ssl`` flag overrides the URL param. Closes #482.
+
 2026/02/09 0.32.0
 =================
 
--- a/docs/run.rst
+++ b/docs/run.rst
@@ -42,6 +42,15 @@ The ``crash`` executable supports multiple command-line options:
 |                               | ``<HOSTS>`` can be a single host, or it can  |
 |                               | be a  space separated list of hosts.         |
 |                               |                                              |
+|                               | Each host may also be given as a full URL,   |
+|                               | e.g.                                         |
+|                               | ``https://user:pw@node:4200/?verify_ssl=0``. |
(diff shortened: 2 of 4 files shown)


amotl (approved):
Thank you for the excellent contribution. Lgtm, but let me also add @mfussenegger for an additional review.

mfussenegger:
I hadn't seen https://github.com/crate/crash/issues/482
Can you elaborate on the motivation for this - why would we need both, `--verify-ssl` and a URL query arg?

florinutz:
we don't *need* both. @amotl wdyt?

amotl:
Hi. I think it is more important for drivers, where you can't define any special flags. However, it is also convenient to be able to use the same style across environments, possibly making the handling easier for scripting purposes. I think GH-482 was written down after a conversation with @WalBeh coming from such spirits. Maybe he can drop a few words about his ambitions?

mfussenegger:
> However, it is also convenient to be able to use the same style across environments, possibly making the handling easier for scripting purposes.

Not sure why - using a driver API and invoking the crash CLI is pretty different.
A few of the CI jobs for CrateDB actually do use a `--verify-ssl False` _and_ use an URL that includes `?verify_ssl=False` because they use both crash and cr8 and there are no issues with that.

Imho the PR isn't needed - more downsides to it due to the additional complexity, both in code but also from a UI perspective for users.

amotl:
Thanks. @WalBeh: Could you elaborate why this was on your wishlist? Maybe it was a misunderstanding: In this case you can blame that on me.

seut:
@WalBeh ping

florinutz:
talked to @WalBeh : closing this

PR #502, opened 2026-06-18
Testing: Use native Testcontainers adapter

### LINKED ISSUE #485: Trim dependencies
## Problem
The project, due to depending on `cratedb-toolkit[testing]`, pulls in a rather large number of dependencies. Please improve.

amotl:
## About
Thanks to @surister and @florinutz, the upstream [testcontainers](https://pypi.org/project/testcontainers/) package will include an adapter for CrateDB. With this patch, the test suite will start using it.

## References
- https://github.com/testcontainers/testcontainers-python/pull/1051
- Resolves: GH-485

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,8 +77,8 @@ setup(
     extras_require=dict(
         test=[
             'crate[test]>=1.0.0.dev2',
-            'cratedb-toolkit[testing]',
             'sqlalchemy-cratedb',
+            'testcontainers[cratedb]>=4.15',
             'zc.customdoctests<2',
         ],
         devel=[
--- a/tests/util.py
+++ b/tests/util.py
@@ -1,5 +1,11 @@
+import logging
+import os
+import sys
+import warnings
+from typing import Optional
 from unittest.mock import Mock
 
+from testcontainers.community.cratedb import CrateDBContainer
 from verlib2 import Version
 
 from crate.client.cursor import Cursor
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -77,8 +77,8 @@ def read(path):
     extras_require=dict(
         test=[
             'crate[test]>=1.0.0.dev2',
-            'cratedb-toolkit[testing]',
             'sqlalchemy-cratedb',
+            'testcontainers[cratedb] @ git+https://github.com/testcontainers/testcontainers-python.git@main',
```
amotl: As soon as there will be a GA release, we can switch over.
```suggestion
            'testcontainers[cratedb]>=4.15',
```

florinutz: should we proceed the same way wherever else we have python testcontainers, I wonder, or wait for GA?

amotl: I think version 4.15 will be around the corner, possibly becoming GA next week already, before the summer heat waves will begin? In this spirit, I think it will be good to wait for an official release before merging this patch.

-- https://pypi.org/project/testcontainers/#history

/cc @alexanderankin

amotl: Hi again. Gentle ping @alexanderankin. It looks like testcontainers v4.15 needs further maturing?

amotl: testcontainers 4.15 was released yesterday. Thank you very much, @alexanderankin!
This spot was adjusted to use `testcontainers[cratedb]>=4.15` now.

tests/util.py
```diff
@@ -47,3 +53,79 @@
...
+
+    @property
+    def http_url(self) -> str:
+        """
+        Return a URL for CrateDB's HTTP endpoint.
+
+        Used to stay backward compatible with the downstream code.
+        """
+        return self.get_http_url()
+>>>>>>> 7ca6fd4 (Testing: Migrate to native `testcontainers` package)
```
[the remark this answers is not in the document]
amotl: That was a merge conflict oversight. Fixed.

setup.py
alexanderankin: > It looks like testcontainers v4.15 needs further maturing?

if you use podman can you check the latest release? somehow it garbles image names (prefix with quay.io or something?) - we had an issue come in and im tempted to just yank

amotl: Hi. We don't use Podman, but Docker on macOS. Apologies.

amotl:
@florinutz: I think the patch is ready to go now.

florinutz:
hey @amotl , thank you for the contribution, first, and sorry for the delayed response!

then, I see you drop the dependency on the toolkit here. My current plan is to have toolkit [updated](https://github.com/crate/cratedb-toolkit/pull/906) first, and only after this to proceed further with any new testcontainers-python work such as this here, or updating the examples

Let me get that released first, then it will unlock using the new testcontainers 4.15 (alongside toolkit and its testcontainer goodies) in the repos so we don't have to reimplement things such as you did here.

amotl:
Hi, and thanks! Please proceed at your disposal.

PR #509, opened 2026-07-07
Mock the connection 

### LINKED ISSUE #508: Nightly tests are failing with connection error
The nightly test run failed with 12 errors (https://github.com/crate/crate-alerts/issues/903), all in `tests/test_commands.py`

```shell
File "/var/lib/jenkins/workspace/CrateDB/nightly/crash_test_nightly/src/crate/crash/command.py", line 441, in _fetch_session_info
    self.connect_info = ConnectionMeta(*self.cursor.fetchone())
TypeError: crate.crash.command.ConnectionMeta() argument after * must be an iterable, not Mock
```

Several unit tests patch `crate.client.connection.Cursor` with a mock but still let `CrateShell()` open a real network connection. Nothing listens on `localhost:4200`, so `_lowest_server_version()` returns `Version("0.0.0")`, `is_conn_available()` is `False`, and `_fetch_session_info()` takes the disconnected branch and the tests pass by accident.

kneth:

Mock the connection instead of the cursor to avoid trying to connect to non-existing server

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #508

### DIFF
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -39,7 +39,7 @@ from crate.crash.commands import (
     ToggleAutocompleteCommand,
     ToggleVerboseCommand,
 )
-from tests.util import fake_cursor
+from tests.util import fake_connect
 
 
 class ReadFileCommandTest(TestCase):
@@ -130,6 +130,7 @@ class ToggleVerboseCommandTest(TestCase):
         self.assertEqual(fake_cmd.reconnect.call_count, 1)
 
 
+@patch('crate.crash.command.connect', fake_connect())
 class ShowTablesCommandTest(TestCase):
 
     def test_post_2_0(self):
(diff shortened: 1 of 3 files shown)


DEVELOP.rst
```diff
@@ -42,15 +42,15 @@ Linting
...
+    isort --check --diff src/crate/ tests/ setup.py
     flake8 crate/crash
 
 Running Tests
 -------------
 
 The tests are run using the `unittest`_ module::
 
-    python -m unittest -v
+    python3 -m unittest -v
```
mfussenegger: Please revert, this seems all unrelated.

mfussenegger (commented):
I suspect this is to fix https://github.com/crate/crash/issues/508 ?

Would be nice of the commit message mentioned that. E.g "Fix commands tests" with a body that mentions that the tests tried to connect to a server. That it replaces one mock with another to fix that is imho secondary.

DEVELOP.rst
kneth: I have reverted it.

bgunebakan (approved):
Thanks!

